PR #151, opened 2014-01-10
Support for partitioning a test suite (to aid in parallelized test runs)

jbergknoff:
Given a big body of features and scenarios, it's nice to be able to partition the list of tests and run those pieces in parallel. One straightforward way to achieve this is: if there are `N` test runners, then the `i`-th runner only runs test scenario `j` if `(j % N) == i`.

This pull request is an implementation of a filter using the modular arithmetic logic described above. This is analogous to the tag filter which determines which scenarios to run based on tags passed at the command line. To invoke a segmented run of the tests, one runs:

`cucumber.js --parts 3 --remainder 0` in runner 0
`cucumber.js --parts 3 --remainder 1` in runner 1
`cucumber.js --parts 3 --remainder 2` in runner 2

### DIFF
--- a/spec/cucumber/ast/filter/modular_partitioning_spec.js
+++ b/spec/cucumber/ast/filter/modular_partitioning_spec.js
@@ -3,18 +3,16 @@ require('../../../support/spec_helper');
 describe("Cucumber.Ast.Filter.ModularPartitioning", function() {
   var Cucumber = requireLib('cucumber');
 
+  var numberPartitions = 4;
+  var remainder = 3;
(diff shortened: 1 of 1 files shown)


sejtan88:
@simondean I didn't manage to use parameters from scenarios to specify which scenarios to run. It always runs all of them. Can you help me with this?

sejtan88:
@simondean I have managed to run test with specific tag using --tags=@abc. But now I have issues with closing browsers. I am closing browsers in hooks, if I put closing browser in After or in AfterFeature, browser is closed, but if I put it in AfterFeatures then it does not, just like that step is somehow skipped. Can you maybe help with this?

Issue #27, opened 2011-09-19
Forgetting to call stepdef callback should fail


aslakhellesoy:
* Go to http://cucumber-js.herokuapp.com/
* Comment out callback() in the Given stepdef
* Run
* Nothing happens

I would expect to see an error message saying "callback not called in /^a variable set to (\d+)$/"

zs-zs:
Hi guys, 

I don't know when do you want to implement the timeout and the promise support, I just needed some quick & nice way to fix these, so I created a small cucumber-js extension called [cucumber-sweetener](https://github.com/zs-zs/cucumber-sweetener) from some of your ideas.

Timeouts can be configured both locally and globally.

It's far from perfect but you can use it as an inspiration.

jbpros:
Thank you @zs-zs, this is a nice intermediate solution! I like small libs! Again, this should be fixed with #187, which is the thing I'm working on atm.

jbpros:
This is somewhat mitigated by the introduction of synchronous stepdefs and promises.

Callback-based step definitions can still make Cucumber die if a callback is forgotten.

jbpros:
Fixed in 2c5cd124724ba128d1d1c2b50ea80f75acef435c

PR #192, opened 2014-06-01
Configurable step timeout

akania:
Based on the discussion in https://github.com/cucumber/cucumber-js/issues/27 i would like to propose this solution :
1) every time step is called ( invoked ) set a timeout with predefined value ( 30s ) , if callback in this step is not called ( success,fail or pending ) call callback.fail() with "Timeout" message - so it is invoked in the scope of the test - this way it has access to all details

2) Regarding configuration - there are two ways :
a) using CLI - add -t ( --timeout ) option with value in ms
b) when constructing new runner ( var runner = Cucumber(featureSource, supportCodeInitializer, options) ) - pass additional 'timeout' option with value in ms
Default value set to 30s is a save bet - in most cases nothing should take more then 30s - but since it is configurable it can be adjusted to the environment  

I believe that the same solution may be applied for hooks. 
I'm not particularly happy with configuration - please review it, i think that there may be better way to access Cucumber.Runtime.stepTimeoutValue...

Being able to define a timeout and get a feedback from the runner that some step has never finished is critical in my opinion - right now i developer may get false-positive results from the tests and it is harder to debug the tests and application code.

Please let me know if is valid solution in your opinion

### DIFF
--- a/lib/cucumber/cli.js
+++ b/lib/cucumber/cli.js
@@ -20,7 +20,7 @@ var Cli = function(argv) {
     },
 
     displayHelp: function displayHelp(callback) {
-      process.stdout.write("Usage: cucumber.js [options] [FILE|DIR]+\n\
+      process.stdout.write("Usage: cucumber.js [options] [[FILE|DIR][:LINE]]+\n\
 \n\
 -r, --require LIBRARY|DIR     Require files before executing the features. If\n\
                               this option is not specified, all *.js and\n\


vkadam:
Is this yet available in laster release? If not whats the ETA?

cades:
+1

jbpros:
This PR doesn't contain anything and is considered *merged* by GitHub (it's not).

I had a look at https://github.com/akania/cucumber-js/compare/timeout -  is that the PR content? If it is, please add scenarios and specs for the new/changed behaviour.

PR #193, opened 2014-06-05
Let step definitions return promises.

### LINKED ISSUE #27: Forgetting to call stepdef callback should fail

samccone:
fixes #173
fixes #27

### DIFF
--- a/lib/cucumber/support_code/step_definition.js
+++ b/lib/cucumber/support_code/step_definition.js
@@ -78,10 +78,21 @@ var StepDefinition = function (pattern, code) {
 
       var parameters      = self.buildInvocationParameters(step, codeCallback);
       var handleException = self.buildExceptionHandlerToCodeCallback(codeCallback);
+
+      var onPromiseFulfilled = function() {
+        codeCallback();
+      };
+      var onPromiseRejected  =  function(error) {
+        codeCallback(error || new Error(StepDefinition.UNKNOWN_STEP_FAILURE_MESSAGE));
+      };
+
       Cucumber.Util.Exception.registerUncaughtExceptionHandler(handleException);
 
       try {
-        code.apply(world, parameters);
+        var result = code.apply(world, parameters);
+        if(code.length < parameters.length && result.then !== void 0) {
+          result.then(onPromiseFulfilled, onPromiseRejected);
+        }
       } catch (exception) {
         handleException(exception);
       }


filipstachura:
Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/cucumber/cucumber-js/193)

Reviewed files:
- lib/cucumber/support_code/step_definition.js @ r1





---

jlin412:
It looks interesting.  Can someone explain to me why we do need this?

johngeorgewright:
:+1:  I would really like this feature.

Selenium webdriver and therefore Protractor use promises. Using cucumber with protractor could be so much easier if promises were handled out of the box.

```js
this.When('I enter "$something" in "$somewhere"', function (text, model, done) {
  element(by.model(model)).sendKeys(text).then(done, done.fail);
});

// compared to...

this.When('I enter "$something" in "$somewhere"', function (text, model) {
  return element(by.model(model)).sendKeys(text);
});
```

jlin412:
That sounds like greAt idea.  Many people are struggling with done callback
when they started.

joshthecoder:
Would be great to get this merged. Right now doing some hackery to patch in promise support by overriding the step methods.

benjamine:
+1

dtinth:
For those who don't want to wait for this pull request to get merged in, you can use [artstep](https://github.com/dtinth/artstep).

It's a more fluent way to write Cucumber.js step definitions.

- For synchronous steps, no need to call `callback()`.
- For asynchronous steps, simply return a Promise or use [ES6 generators](https://www.npmjs.com/packages/co). Which means it also works with:
    - [ES7 async functions](https://github.com/tc39/ecmascript-asyncawait)
    - [asyncawait](https://github.com/yortus/asyncawait)

jlin412:
@samccone and @jbpros What is reason of delaying this feature?
This feature will make cucumber.js so much easier to use.

bmustiata:
If you want to add support for promises and steps without callbacks you can also use [cucumber-promise](https://github.com/bmustiata/cucumber-promise) ([npm link](https://www.npmjs.com/package/cucumber-promise)).

The implementation of [promises](https://github.com/bmustiata/cucumber-promise/blob/master/features/promises.feature) and [steps without callbacks](https://github.com/bmustiata/cucumber-promise/blob/master/features/no-callbacks.feature) are tested with cucumber.js itself, to guarantee it plays nice with your existing tests, and is minimally invasive (just wrap the steps function into one function call).

jlin412:
I like this

jbpros:
Merged in as https://github.com/cucumber/cucumber-js/commit/272dc5e64b4f0e3c9e6704a11b93e9e7cf519906. Released as of 0.5.0.

samccone:
:+1:

Issue #157, opened 2014-02-27
Extreme stack traces


TheLudd:
I am just trying out cucmber-js and I notice that when a test fails, an enormous stack trace is produced (160 lines). This makes it really hard to understand which test failed and why. Is there some kind of workaround or solution to this problem?

cades:
Mocha encounter the [same problem](https://github.com/mochajs/mocha/issues/545), and there's a nice [solution](https://github.com/keyvanfatehi/js-simple-stacktrace) using pipe to shorten stack trace.

dtinth:
I think I just found a gem. [clarify](https://www.npmjs.com/package/clarify) is a module that removes Node.js-related stack frames from the stack trace. Under the hood, it uses [stack-chain](https://www.npmjs.com/package/stack-chain) module, which allows stack frames to be extended, filtered, and replaced.

Therefore, I created a package `hide-stack-frames-from`, and you can use it like this:

```js
require('hide-stack-frames-from')('cucumber')
```

jlin412:
Thai, How about submitting a PR using this code to give an option to hide
node.js stack trace?

dtinth:
@jlin412 I am already happy putting that line of code in some `support` file. Therefore, although it doesn't hurt to do so, I don't see the need of it being in Cucumber's core.

I also use it to hide stack frames from Bluebird, and etc. That's why I prefer being able to configure all of them from a single supporting script.

sebs:
```
require('hide-stack-frames-from')('cucumber')
```

Works like a charms and fixes my problem. thanks

plato-cambrian:
thanks @pumbers !

floribon:
Just a heads up, not sure if there is a better solution today on March 2016.

I tried using `hide-stack-frames-from` but it was failing on some scenario, and would leave some node related traces.

Instead I used @laurelnaiad with the addition of the error message. Indeed the snippet was removing any custom message sent by `throw new Error` or a failing promise. However some cucumber error messages were still ugly, such as:

```
UnknownError: unknown error: Element is not clickable at point (580, 60). Other element would receive the click: <div class="collapse navbar-collapse">...</div>
        (Session info: chrome=48.0.2564.116)
        (Driver info: chromedriver=2.21.371459 (36d3d07f660ff2bc1bf28a75d1cdabed0983e7c4),platform=Mac OS X 10.11.3 x86_64) (WARNING: The server did not provide any stacktrace information)
[...]

PR #227, opened 2014-08-14
Adding short stack traces as an option

AndrewJohnBenjamin:
Updated the package to include a command line option that allows the user to specify whether or not they would like single line stack traces upon step failure.

Currently the stack traces that are output by the package in the event of a step failure are pretty large and make it a little more difficult to see what has failed. The changes I have made cut this trace down to a single line and shows only the line in the code that caused the step to fail.

The changes involved slight modifications to runtime, configuration, argument parser and step_definition.

I believe this addition would benefit others as it seems this issue has been flagged already.

https://github.com/cucumber/cucumber-js/issues/157


jbpros:
Thanks @AndrewJohnBenjamin for this.

The short stack traces are now implemented (see f584233d5a6f13bd3dc380e0e622c5baea74f438).

There are several other features in this PR. Please open new, separate, PRs for those.

PR #228, opened 2014-08-15
Internet Explorer compatibility fix and bugfix

jason0x43:
* IE9-10 require explicit contexts to be passed to function pointers in certain cases (http://msdn.microsoft.com/en-us/library/ie/gg622930%28v=vs.85%29.aspx). The `process.nextTick` implementation used for IE, which is based on `setImmediate`, was updated to accept and pass an explicit context to the callback.
* The implementation for DataTable.Row did not declare `self` to be a local variable, and IE9 wouldn't allow a new value to be assigned to it.

### DIFF
--- a/lib/cucumber/ast/data_table/row.js
+++ b/lib/cucumber/ast/data_table/row.js
@@ -1,7 +1,7 @@
 var Row = function(cells, uri, line) {
   var Cucumber = require('../../../cucumber');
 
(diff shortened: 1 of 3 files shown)


jbpros:
I'm closing this for now, as I can't reproduce the problem. Was it fixed by browserify?

Please reopen if needed.

Issue #282, opened 2014-12-18
Attachments are not cleared in before/after hooks


nebehr:
If an attachment is created in after hook, it gets embedded to all subsequent after hooks of the current scenario and all before hooks of the next one. This is because attachments are cleared in AstTreeWalker.witnessNewStep (https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/runtime/ast_tree_walker.js#L223) but it does not get called for before/after hooks.

To reproduce create a feature file with 2 scenarios, define before and after hook and make an attachment in the latter:

```javascript
this.Before(function (callback) {
      callback();
});

this.After(function (scenario, callback) {
        scenario.attach('BLABLABLA ', 'text/plain');
        callback();
})
```

When this feature is run with JSON formatter, total number of embeddings will be 3 instead of 2.

PR #284, opened 2014-12-22
Fix for issue #282

nebehr:


jbpros:
Merged in as 75822d8.
Thanks @nebehr.

PR #287, opened 2014-12-24
A new feature for switching environment of data table. 

puritys:
We used cucumber to test our websites and met a problem about the data table of scenario outline features. 
Data table will test all data rows which we defined, but we need to skip some data rows when these testing are running in different environment, such as alpha, beta, production.

Here is the example of new feature.
<pre>
     Feature: testing scenarios
        Scenario Outline: scenario outline 1
          When step <id>

        Examples:
          |_env   | id |
          | alpha | a  |
          | beta  | b  |
          | all   | c  |
</pre>
If you executed command cucumber.js append with "-data_env beta", then cucumber will only execute two data rows, both of them are "beta" and "all". 
<pre>
jasmine-node spec && ./bin/cucumber.js -i features/xx.feature  -data_env beta
</pre>

### DIFF
--- a/features/scenario_outlines.feature
+++ b/features/scenario_outlines.feature
@@ -98,3 +98,24 @@ Feature: Scenario Outlines and Examples
     And the step "step b" passes
     And the step "step c" passes
     And the step "step d" passes
+
+  Scenario: Data table can switch environment by _env tag
+    Given the following feature:
+      """
+      Feature: testing scenarios
+        Scenario Outline: scenario outline 1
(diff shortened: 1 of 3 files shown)


jbpros:
Thank you @puritys.

Cucumber doesn't (want to) know about environments nor the content of examples data (even through some underscore-prefixed column).

I think what you're looking for is the currently missing #305.

Please reopen if you don't agree.

A PR on #305 is more than welcome!

Issue #307, opened 2015-02-20
npm Node Version Warning on Install


dschinkel:
Just wondering if this is anything I need to worry about and what exactly this is saying with the version mismatch here.

Note: I'm using io.js if that helps...
![screen shot 2015-02-19 at 10 59 44 pm](https://cloud.githubusercontent.com/assets/338798/6281708/9262388a-b88b-11e4-87e9-6309a38ee9aa.png)

samccone:
yes that would be the issue @dschinkel :)

dschinkel:
What is, that Cucumber.js isn't compatible with io.js?

samccone:
oh that cucumber-js simply has not been run on io.js yet
I see no reason why it would not be compatible, we simply need to update the package.json and update the test suite to test against io

dschinkel:
What does "updating the test suite" specifically involve?

dschinkel:
When can we get this updated then??

aslakhellesoy:
Io.js is fairly new, but I'm sure there's people who are running Cucumber.js on it already. Has anyone experienced any errors? (I'm talking about errors, not warnings).

dschinkel:
Yea sorry, just anxious to get started...:)

dschinkel:
I'm running io.js 1.3 now just for more info...

samccone:
@dschinkel @aslakhellesoy I have fixed the issue, as expected io works just fine

#308

dschinkel:
That is awesome thanks lot for getting to this!!

- Dave

dschinkel:
I ran the same exact command npm install -g cucumber.  Has your change been pushed to the npm registry or whatever?  Is this the command you tested for the fix?

samccone:
it has not been merged, the PR is still open, waiting on @jbpros

dschinkel:
I suppose duh I can just update my own package.json.

samccone:
sure you can point your cucumber_js install at my commit

`npm i cucumber/cucumber-js#0f13417bfecfa0dcbdac16741816bfda84d7d536 --save`

https://docs.npmjs.com/cli/install

eduardoinnorway:
Well the dependency jsdom is io.js only, so you should update your read me, because your getting started tuts don't work anymore with Node.

PR #308, opened 2015-02-22
Add io.js to grid

samccone:
Via #307 

Since io.js follows semver, an install of cucumber-js would complain due to the version of io not matching the range of node specified in engines.

This should resolve the issue.


samccone:
@jbpros not sure why circle is running but... the tests are passing

jbpros:
Merged as 89088a5d94e5b6ffa4a3b0eed6175409dd94273b

PR #257, opened 2014-10-05
Allow Cucumber to load custom formatter

afternoon:
@joshtombs PR #215 + the requested additional tests.

Issue #90, opened 2012-10-02
Reporting - Output to file


peteclark82:
It is currently not possible to specify an "output" file as a CLI argument for saving reports. The JSON formatter seems to write directly to STDOUT, which can be a problem when debugging (using console.log) or if the program fails, producing invalid JSON.

It would be desirable if this worked like cucumber in ruby, allowing multiple formats saved out to separate output files.

cedriclombardot:
Hi,

Anyone can help me, i've not found how to configure properly json output into file.
I use gulp+protractor+cucumberjs

Thanks in advance

jlin412:
@cedriclombardot 

There is a way to select a directory to json output from protractor from 1.5.0 though I haven't used it.
Currently, I added a hook so I can use pretty and json formatter together:

``` javascript
module.exports = function JsonOutputHook() {
  var Cucumber = require('cucumber');
  var JsonFormatter = Cucumber.Listener.JsonFormatter();
  var fs = require('fs');

  JsonFormatter.log = function (json) {
    fs.writeFile('test/e2e/output/cucumberReport.json', json, function (err) {
      if (err) throw err;
      console.log('json file location: test/e2e/output/cucumberReport.json');
    });
  };

  this.registerListener(JsonFormatter);
};
```

cedriclombardot:
I've not seen the `resultJsonOutputFile` seems to works magickly and ready
to go in xml with : https://www.npmjs.com/package/protractor-cucumber-junit

Cédric



2015-01-02 18:34 GMT+01:00 Jason Lin <notifications@github.com>:

cades:
+1

fraserc:
So could we please get an update on when this will get committed to master? I'm currently using cucumberjs minus protractor (or any other selenium style web test framework) so I need cucumberjs to produce a formatted report.

auranine:
Any update on this? This has caused us enough issues to think about patching this module or looking at parallel-cucumber (which handles this case).

robcleghorn:
+1

charlesh21:
I follow this description to save cucumberjs json format.

https://github.com/mrooding/gulp-protractor-cucumber-html-report#saving-cucumberjs-json-to-disk-when-using-protractor

1. Add a listener to the CucumberJS JSON formatter and save it to a file
2. protractor.conf.js add a reference to the hook listener

It's work

noahehall:
if anyone is missing happy hour due to cucumber js reports:

[...]

PR #309, opened 2015-02-26
custom formatter

samccone:
Here we are, custom formatters! rebased on top of master, tests passing and docs. Lets ship this sucker!

Via #90 
Via #188 
Via #215 
Via #257


samccone:
ping @jbpros

stjohnjohnson:
Please yes! :shipit: :sheep: :coffee: 
nod

afternoon:
:+1:

stjohnjohnson:
So, who are we waiting for?

petey:
Yes! :shipit:

mal thumb

samccone:
Just waiting on some eyes from @jbpros

mattwynne:
My one note of caution with this is the experience I've had with having 3rd party code depending on the Ruby Cucumber's original formatter API, which was very, very complex.

I can't see from this PR what the formatter API is, but just bear in mind that once it's out in the wild you'll find it much harder to change it. So make sure you're happy with it now.

samccone:
heh yeah it is not exactly "simple" to use right now... but..... we can follow semver when we change it so it should not be such a big issue.

jbpros:
This is great stuff! Thanks @samccone.

Also, thanks @mattwynne for the note of caution, it makes sense.

3rd-parties should rely on the listener API, based on events, that allows for pretty good decoupling between the internals of Cucumber and formatters/listeners. 

I think this PR lacks two things:

1. Documentation on how to write *good* plugins (i.e. do not rely on internals, use events!). Maybe an example?
2. Some gherkin scenarios testing this e2e.

mattwynne:
Could 1&2 not be combined @jbpros? i.e. use the Gherkin scenario as the documentation?

Crazy talk, I know...

stjohnjohnson:
@samccone Can I help with the docs or gherkin?

charlierudolph:
@samccone this looks pretty good as is but is still missing documentation and feature tests. I can help with those if you are okay with me pushing to your branch.

spec/cucumber/cli_spec.js
```diff
@@ -35,30 +35,25 @@ describe("Cucumber.Cli", function () {
     describe("when the help is requested", function () {
       beforeEach(function () {
         configuration.isHelpRequested.andReturn(true);
+        cli.run(callback);
```
charlierudolph: Please revert the changes to this file as they are unrelated

jbpros:
I second @charlierudolph on scenarios and docs for this.

afternoon:
@charlierudolph @jbpros I added documentation to `--help` and to `README.md` here: https://github.com/cucumber/cucumber-js/commit/053ddbc1f1e0ae74167a69ee594fc833106d5f31

Do we need to expand these, or add other documentation elsewhere?

charlierudolph:
The documentation I'm talking about is how to build a formatter.
* what options are passed to the constructor 
* what methods it can or should implement and the arguments those methods are passed

afternoon:
Sounds so obvious when you say it. :disappointed:

Is there a single point of reference for options, or do we need to infer them from the code, e.g. [formatter.js](https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/listener/formatter.js), [summary_formatter.js](https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/listener/summary_formatter.js) and the [other listeners](https://github.com/cucumber/cucumber-js/tree/master/lib/cucumber/listener)?

charlierudolph:
The options for the constructor come from [here](https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/cli/configuration.js#L11). The coffeescript snippets option needs to go away. 

I think we probably want to suggest inheriting from Cucumber.Listener.Formatter. Might be good to just give people a skeleton for a formatter.
```js
function MyFormatter(options) {
  var Cucumber = require('cucumber');

  if (!options)
    options = {};

  var self = Cucumber.Listener.Formatter(options);

  self.handleBeforeFeaturesEvent = function handleBeforeFeaturesEvent(event, callback) {
    var features = event.getPayloadItem('features');
    callback();
  };

  self.handleBeforeFeatureEvent = function handleBeforeFeatureEvent(event, callback) {
    var feature = event.getPayloadItem('feature');
    callback();
  };

  self.handleBeforeScenarioEvent = function handleBeforeScenarioEvent(event, callback) {
    var scenario = event.getPayloadItem('scenario');
    callback();
  };

  self.handleBeforeStepEvent = function handleBeforeStepEvent(event, callback) {
    var stepResult = event.getPayloadItem('step');
    callback();
  };

  self.handleStepResultEvent = function handleStepResultEvent(event, callback) {
    var stepResult = event.getPayloadItem('stepResult');
    callback();
  };

  self.handleAfterStepEvent = function handleAfterStepEvent(event, callback) {
    var step = event.getPayloadItem('step');
    callback();
  };

  self.handleAfterScenarioEvent = function handleAfterScenarioEvent(event, callback) {
    var scenario = event.getPayloadItem('scenario');
    callback();
  };

  self.handleAfterFeatureEvent = function handleAfterFeatureEvent(event, callback) {
    var feature = event.getPayloadItem('feature');
    callback();
  };

  self.handleAfterFeaturesEvent = function handleAfterFeaturesEvent(event, callback) {
    var features = event.getPayloadItem('features');
    self.finish(callback);
  };

  return self;
}

module.exports = MyFormatter;
```
Right now the objects are based on the internals of cucumber and probably need to be wrapped up somewhere so we can offer a more consistent interface as the internals are changing in a big way when we update to gherkin3. Probably need something similar to the [wrapper](https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/api/scenario.js) around a scenario that is passed to hooks, for each of the objects. @jbpros what are your thoughts?

jbpros:
I agree. Event objects shouldn't be exposed to the outside world (the same holds true for event listeners in userland, btw).

I'm thinking we could use the simpler emitter API for listeners:

```javascript
function MyFormatter(suite) {
  // suite is an emitter instance injected by Cucumber
  suite.on('beforeFeature', function (feature) {
    // do your stuff
  });

  suite.on('beforeScenario', function (scenario) {
  });

  suite.on('afterScenario', function (scenario, result) {
    // receives the hypothetical result object
  });
  
  suite.on('beforeStep', function (step) {})
  suite.on('step', function (step, result) {})
  suite.on('afterStep', function (step) {})

  // etc.
}
```

This is only a listener. Shouldn't we use the name "listener" instead of "formatter" in this new API?

Getting back to the "payload" objects, with such an emitter, each event is accompanied with one or more things and doesn't need a `getPayloadItem` method anymore.

I'm also tempted to rename the events to something like `before-scenario` or even `before scenario`.

mattwynne:
FWIW, we've got with `before/after_test_case` for as opposed to `before/after_scenario`. We wanted to make a clean break from any confusion with the old AST-based formatter API.

ntosicct:
Any updates on this?

a11smiles:
when will this be working!?!?

mattwynne:
👍

a11smiles:
@mattwynne any time expectations around this?  I need to create TeamCity and AppVeyor formatters for CI.

mattwynne:
What are the next steps @jbpros? Do we still need some docs? Maybe you could help with that @a11smiles?

a11smiles:
@mattwynne I'm available to do so...indeed, what are the next steps @jbpros ?

kchokhawala:
@jbpros: Is there any update on this ? Is this available to use ? I see there have been some conflicts. I would like to use this, if this is good to (or as and when) use, please provide link to some finalized documentation. Thanks.

charlierudolph:
@kchokhawala this isn't ready. I'm working on some sizable changes to the formatters. Once those are done we can circle back to custom formatters. I'm going to go ahead and close this due to its age. Issue held in #516

Issue #250, opened 2014-09-27
scenario does not have an exception property like ruby version


jlin412:
I am trying to write an after hook.  When there is a particular error message, I will do some cleanup steps.

This is what I am trying to write:

``` javascript
  this.After(function (scenario, done) {
    if (scenario.isFailed()) {
      if(scenario.exception.message.indexOf('UnexpectedAlertOpenError') > -1) {
        browser.switchTo().alert().accept();
        browser.sleep(3000).then(done);
      }
    } else {
      done();
    }
  });
```
https://github.com/cucumber/cucumber-js/blob/master/lib/cucumber/ast/scenario.js
scenario.exception.message is the ruby version of scenario object but I can't seem to find any equivalent to find error message from scenario.js.  I have a workaround to get error message but I think scenario.js should include a method or property to get error message.

jbpros:
Closed by a773aa2

PR #317, opened 2015-03-31
Add failure exception to scenario object

ertrzyiks:
As @jlin412 suggested in #250 it would be nice to have access to thrown exception in scenario object.


sskorc:
:+1:

karolsojko:
:+1:

ertrzyiks:
Looks like latest npm doesn't work with node 0.8

jbpros:
Merged in as a773aa2

PR #336, opened 2015-05-15
Update progress_formatter.js

JBeen:
Added information about step in log while debugging tests


jbpros:
The progress formatter is supposed to be concise, usually running on a CI server. This clutters the output and doesn't follow the usual "progress" formatter output you see in other cucumber implementations and other testing tools.

Isn't the pretty formatter what you're looking for?

PR #354, opened 2015-06-20
normalize OS-specific path separators when performing assertions on cucumber output

### LINKED ISSUE #353: End-to-end tests are failing on Windows
End-to-end tests are failing on Windows because the output of the pretty formatter contains feature file paths and these paths contain OS-specific directory separators. Under Windows it's backslash while the expected output contains slashes (the default separator in Unix).

We could fix this in two ways, we could either fix the actual formatter code to output paths consistently under different operating systems or we can fix the assertion code to ignore the separators.

I think it's more feasible to rely on the OS setting and fix only the assertion code.

zs-zs:
this pull request fixes #353


jbpros:
Merged in as ed4f769e09f58761f0cb9266d4b703c5f2344bfd

PR #358, opened 2015-06-23
Fix ruby tests on Windows

### LINKED ISSUE #357: Ruby tests are failing on Windows
Yet another issue with shebangs which are not working under Windows. We have to call node manually. The issue is similar to #353, but this one is more cumbersome to fix, because aruba's command line executor is directly called from the steps.

zs-zs:
This pull request fixes #357


jbpros:
Thanks @zs-zs, merged in as ef99935

PR #359, opened 2015-06-23
Step definition promise check

### LINKED ISSUE #347: Steps with no arguments will fail in v0.5.0
Hi guys,

I've just realized that the way you implemented promise support in v.0.5.0 will cause steps with no arguments fail.
So I cannot write steps like this anymore:

    When('..pattern...', function() {
        var firstArg = arguments[0];
        ....
    });

You make a check for code.length, the length of the argument list of the "code" function:

    +        if (code.length === parameters.length - 1 && result.then !== void 0) {
    +          result.then(onPromiseFulfilled, onPromiseRejected);
    +        }

I would just check for the existence of the then function instead. It would be enough in most cases.

The biggest disadvantage of the current solution is that it prevents the implementation of generic step definition APIs based on Cucumber-js. (like my small lib [cucumber-sweetener](https://github.com/zs-zs/cucumber-sweetener))

zs-zs:
This pull request fixes #347

### DIFF
--- a/features/step_definition_callbacks.feature
+++ b/features/step_definition_callbacks.feature
@@ -26,3 +26,9 @@ Feature: Step definition callbacks
     When Cucumber executes that mapping
     Then the mapping is run
     And the scenario passes
+
+  Scenario: succeed with promise if implicit parameters
+    Given a promise-based mapping with implicit parameters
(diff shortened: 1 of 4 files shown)


jbpros:
Merged in as b138589

PR #361, opened 2015-06-26
Update walkdir module to 0.0.10

repko-artem:
There's a bug in walkdir <0.0.10 in Windows (ntfs). Step definitions and features in different files have the same inode attribute. It causes that some features and some step definitions don't load, when tests run.


jbpros:
Thanks @repko-artem. Merged in as bf039e31a2071570b37b1702a8e788cfe83a581e.

PR #368, opened 2015-07-09
Fix World code examples in README.md

renegare:
Hi,

Was stuck with the examples given on the README for moment. I believe it needs to be updated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ function World(callback) {
     callback(); // tell Cucumber we're finished and to use 'this' as the world instance
   };
 }
-module.exports.World = World;
+module.exports = World;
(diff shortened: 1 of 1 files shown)


samccone:
hi @renegare did this change work for you?

renegare:
Nope :) but either way I think something needs to change. best to close this?

samccone:
yeah lets close this, but feel free to open an issue describing this problem so that we can put in a real fix

thanks @renegare :clap:

jbpros:
It's not recommended to expose the World constructor at the root of a module as it might be executed by cucumber's support code autoloading. The current README looks good to me.

PR #482, opened 2015-12-22
Consistent casing for headers

kevgo:
Most headers have only the first word capitalized. This PR changes the remaining few headers that have every word capitalized to be consistent with that.


jbpros:
Merged in as 6ede7647431b3c7133eb0f8de8da23c3af4751ed

PR #483, opened 2015-12-22
Consistently uppercase syntax for "Cucumber"

kevgo:
Cucumber and Cucumber.js are names and should therefore be uppercased.


jbpros:
Merged in as 0b0b1f72ef3202f6f003632cfdbfc4ff186248a9

PR #484, opened 2015-12-22
Remove single-use acronym

kevgo:
SUT is used only once in this document, so it can (and should) be omitted. It's also not really necessary here, since this document doesn't compare testing architectures, but more how to set up a particular testing utility.


jbpros:
Merged in as 6eb39ebb4979b406e359b26bb3a25316e298a4cf

PR #486, opened 2015-12-22
Improve wording

kevgo:
"so-called" sounds a bit derogatory and insecure around how Windows file handling works. I think the sentence reads better without it.


jbpros:
Merged in as 1e17625793b418328ed98c5fe8fbf07839f35203

PR #510, opened 2016-02-04
Added World constructor breaking change to documentation

mrwithersea:
The update to the documentation isn't adequate for a public api that has so many examples and implementations that make use of the asynchronous world constructor. As this breaking change is difficult to hunt down in the release notes I have added it to the README


jbpros:
Merged in as c884dda

PR #566, opened 2016-04-26
Correct tagged hook example

cbrunnkvist:
The way it is currently implemented we should pass a variable amount of string parameters rather than an object with `tags` array .

This updates the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,8 @@ Hooks can be conditionally selected for execution based on the tags of the scena
 // features/support/hooks.js (this path is just a suggestion)
 
 var myHooks = function () {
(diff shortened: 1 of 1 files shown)


cbrunnkvist:
_NOTE:_ Please disregard - I just realized that the project I was working on depended on cucumber-js __0.9.5__ 👎 and I can also see that the change was [properly documented in the changelog](https://github.com/cucumber/cucumber-js/blob/master/CHANGELOG.md) as of version __0.10.0__. 

I was only fooled to believe that the error was in the docs because the the result was an uncaught exception.

PR #853, opened 2017-06-11
WIP: Fix for #846, Support code aliases break step definition stacktrace (v2.3.0)

pittgoose:
This is a WIP PR for fixing issue #846 with @charlierudolph


charlierudolph:
Please remove the changes to the `package.json` file for this PR. As part of the release process, the version will be updated and new contributors will be added to the `package.json` file. Also do you have any attempts to solve this already or any specific questions to get started?

pittgoose:
@charlierudolph I reverted my changes to package.json.
I tried digging into this but the stacktrace was entirely in node_modules. I was not able to figure out how to find a stacktrace which contained the proper path to the step definition file. Are you familiar with the code?

charlierudolph:
Were you looking at this [function](https://github.com/cucumber/cucumber-js/blob/5c4bb4353fc10121e271f79d01fb5474cf7b4e76/src/support_code_library/helpers.js#L47)? That gets called in defineStep / defineHook / registerHandler. We may be able to write a failing feature test for this or even a unit test. Creating a failing test case is probably the best case to start

pittgoose:
Thanks I'll take a look

pittgoose:
@charlierudolph I've added failing tests now for defineStep / defineHook / registerHandler

charlierudolph:
@pittgoose so I believe we need to update the function I linked in the previous [comment](https://github.com/cucumber/cucumber-js/pull/853#issuecomment-307835465). That is currently assuming the user code is three up the stack. However, it now may be four up the stack when using the direct imports. In order to combat this, I think we should look through the stack searching for the first one that is not in the cucumber node module. [Here](https://github.com/cucumber/cucumber-js/blob/d1dd18965c2c2b7da147814536f123ee6b938ef0/src/cli/install_validator.js#L7) is an example of getting the project path which we would need to check if the stack contains in order to tell if it in the cucumber node module.

pittgoose:
@charlierudolph could you let me know if you have some time to work on this at the same time? I'm totally confused but would really like to fix this.

charlierudolph:
superseded by #881

PR #1114, opened 2018-08-09
727 - CLI option --retry

hurrikam:
This branch implements the CLI `--retry` flag to immediately re-run a failing scenario. 
If omitted the flag defaults to 0 (no retry logic).

A positive number will cause the `test_case_runner` to re-run its scenario up to the specified number of times if the previous run ended with `result.status === Status.FAILED`.
Any different result will immediately pass through (either the test case succeeded or it was marked as skipped or ambiguous).

That means the retry logic applies only to genuine failures of the scenario.

From the event protocol point of view the `test-step-started` event will be emitted for retrial steps as well, so the console will print more `.` than the actual length of the scenario, when retrying.

On the other hand, the `test-case-started` and `test-case-finished` events will be emitted just once (even if the test case has been retried).
This ensure flaky test cases won't appear more than once in formatter outputs (like the JSON one), thus affecting the total count of test cases reported per scenario/suite. Only the last run of test will be logged.


VadimZhiltsov:
I vote for this changes. Nice work!

jain-neeeraj:
Dear Reviewers,  
Can you please review this as it is one of the most crucial feature.  It is a must have thing to develop any E2E test cases.

charlierudolph:
How do formatters handle this? For this to work, I think we would need a new status that is "failed, will retry" so the formatters can give warnings about this.

Also please remove the package-lock.json. This repo uses yarn instead of npm

hurrikam:
@charlierudolph as explained in the first comment the test case is reported to start and finish only once (even if it will be re-run), so at the moment the feature is transparent to formatters. Doing otherwise would require changes in existing formatters which I think should be discussed first.
This can be seen as the base implementation of the feature which is compatible with existing formatters but which can be later changed.

charlierudolph:
I don't like that it is transparent to formatters though. I'd prefer people have some way to view the failed scenarios that passed on retry if they want. This could be a formatter option, `showRetryFailures`. If its set it displays the errors with a note that it is going to be retried.

If anything passes after a retry we could give a message: 

```
X scenarios passed after retry, use `--formatOptions '{"showRetryFailures": true}'` to view the failures
```

I'd also like to explore making this more granular so you signify specific scenarios that should be retried instead of having all scenarios be retried. That can be dealt with separately but something to think about.

hurrikam:
If we want existing formatters to report all test runs without any breaking change, we could just report every test run but then that:

1) would make it look like your feature file had more test cases than expected
2) a failed test would mark the feature as failed (even if passing on a second attempt)
3) if the fail fast flag is specified then test execution would immediately terminate with a failure

As per the granularity we may re-run only scenarios which are marked with an ad hoc flag (i.e. @retry)
However, in our real life case any tests in our suite may need to re-run because the issue is with the device itself being flaky at some points. So there should still be the possibility to have a global value without having to mark all the scenarios explicitly.

charlierudolph:
> would make it look like your feature file had more test cases than expected
> a failed test would mark the feature as failed (even if passing on a second attempt)
> if the fail fast flag is specified then test execution would immediately terminate with a failure

If we introduce a new status that can be properly handled by the rest of the system, I don't think we will hit any of these issues. For reporting the failures that need to retried we just need to ensure its clear that these failures will be retried.

> So there should still be the possibility to have a global value without having to mark all the scenarios explicitly.

I'm fine with keeping a global value, but I want a way to turn this on per scenario as well so your test run doesn't spend time rerunning failed scenarios that shouldn't be retried.

chipironcin:
I am keen to help with this feature as it is also crucial to the work I am currently doing (not just retry because my test is flacky but because some other dependency out of our reach).

Let me know if I can help with something (maybe testing)

chipironcin:
If I read well, only the following bits are missing:
1. Control per scenario retry
2. Control aggregation on reports

For the first one, we might need two different options for the command line: `retry` and `retryTag`.
The official implementation of Cucumber in Ruby has `retries` as retries for all scenarios, so if we want to match we would need `retryTag` for enabling specific tag retries.

For the second one, I think the main Ruby implementation just throw them all (failure and pass) into the report. We could have a flag controlling that as mentioned here: http://mkolisnyk.github.io/cucumber-reports/failed-tests-rerun

WDYT?

Links:
https://github.com/cucumber/cucumber-ruby/blob/c45aa7fcb3441cd19c2a99817c9afbdb24656ec2/features/docs/cli/retry_failing_tests.feature
http://mkolisnyk.github.io/cucumber-reports/failed-tests-rerun

src/cli/argv_parser.js
```diff
@@ -111,6 +111,18 @@ export default class ArgvParser {
...
         []
       )
+      .option(
+        '--retry <NUMBER_OF_RETRIES>',
+        'specify the number of times to retry failing test cases (default: 0)',
+        parseInt,
+        0
+      )
+      .option(
+        '--retryWithTag <NUMBER_OF_RETRIES>',
```
charlierudolph: I think a better option would be `--retryTagFilter <TAG_EXPRESSION>`, specify what subset of scenarios can be retried.

hurrikam: Makes sense. In that case the --retry option is still needed to specify the count of retrials, as I guess it can't or shouldn't be part of the tag expression?
If that's the case, what's the behaviour when --retryWithTag is specified but --retry is omitted?

chipironcin: `--retry 4` -> Retries everything up to four times
`--retryTagFilter "@retry"` -> Does not retry anything
`--retryTagFilter "@retry" --retry 4` -> Retries @retry tests up to four times

hurrikam: Yes, that's what I thought. However I would like the CLI to throw an error if user specifies --retryTagFilter without --retry itself. I don't like when CLI parameters are silently ignored.

charlierudolph: Sounds good on throwing an error if using retryTagFilter without retry

hurrikam: Ok, I started to make those changes. Quite busy these days but I will try to do some work over the weekend. :)

hurrikam:
Hi guys,
after the last changes this is the current behaviour of the `retry` feature:

1. `--retryTagFilter` is implemented as discussed, whit the following behaviours:
a) if specified in addition to `--retry` will run only feature/scenarios matching the given tag expression
b) CLI will throw an exception if the flag is specified but `--retry` is missing (or is set to zero)

A consideration here: the CLI currently allows options like `--parallel` and `--retry` to be set to 0. I think there should be a check to ensure they're positive values otherwise there's no point in setting them.
I've kept the existing behaviour to make sure the PR won't create any break change, but then throwing an error in the specific case where `--retryTagFilter` is used with `--retry` set to 0 may feel inconsistent?

2. A new status code `FLAKY` was introduced. Then a test passes after retries then:
a) The test cases is marked as FLAKY but CLI still exits with error code 0 (if no other failures)
b) The `Summary formatter` will report the total count of flaky tests
c) The `Summary formatter` will print flaky test cases under the `Warnings` section (but no details on the failing steps)

Do we think it makes sense to mark individual steps  as `FLAKY` (steps which will be re-run in the following retry)?
If we go that way the `Summary formatter` in example could still print the test case just once but highlight all flaky steps in a different color.
Similarly the `Progress formatter` could print in example '....R' rather than '....F' when a step fails (but it will be re-tried). 

All the feedback is welcome

src/runtime/test_case_runner.js
```diff
@@ -167,24 +169,38 @@ export default class TestCaseRunner {
...
+      }
+      if (retries < this.retries) {
+        delete this.result.exception
+        this.result.status = Status.PASSED
+        this.testStepIndex = 0
+      }
+    }
+    if (this.result.status === Status.PASSED && retries > 0) {
+      this.result.status = Status.FLAKY
+    }
```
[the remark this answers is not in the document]
hurrikam: @charlierudolph it seems like my reply was not published, so I'll rewrite it :)

I remember that firing the `test-case-finished` with a `FAILED` status would immediately stop the test, without a chance to re-run it. Having the `FLAKY` status should help now.

From the reporting point of view it still remains the question that formatters (and any report generated out of them) would display multiple runs of the same test case, without the capability of aggregating them.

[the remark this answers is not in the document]
hurrikam: I have moved the two events inside the loop and everything seems fine so far. Just reviewing unit tests for the test case runner.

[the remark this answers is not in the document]
charlierudolph: > I remember that firing the test-case-finished with a FAILED status would immediately stop the test, without a chance to re-run it.

That's why I was suggesting a RETRY status (but flakey could work as well). But I want two test cases instead of just one so the formatters can differentiate the runs.

> From the reporting point of view it still remains the question that formatters (and any report generated out of them) would display multiple runs of the same test case, without the capability of aggregating them.

Yes, formatters would need to deal with the multiple runs and aggregate them if they choose or only display the last run of the test case (which I think is what would happen currently)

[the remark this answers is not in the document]
hurrikam: @charlierudolph once I fix the tests I will have a look into the formatters behaviour as well

hurrikam:
With the latest changes this is the current behaviour of the `retry` feature:

1. When a test case run fails but will be retried, `testCase.result.status` is set to `FLAKY` and `test-case-finished` is fired with that status
2. After the last run, the test case is marked either `PASSED` or `FAILED` to reflect the final outcome

As per formatters:

1. `Summary` formatter prints a test case which passed after retries in the `Warnings` section
2. `Progress Bar` formatter prints every run of a test case which will be retried
3. `JSON` formatter includes the `retryAttempt` property in scenario data, if any retry occurred
4. `Progress` formatter prints a character for every step of each run. It still uses 'F' for failing steps which will cause a retry of the test case.

charlierudolph:
Cool. Thanks for the updates. I'll take a deeper look at this week and hopefully we can get it released!

hurrikam:
@charlierudolph thanks. Just a couple of notes:

1. If we want the capability to completely omit any information about retries in formatter reports, then we still need to add the `--formatOptions` parameter you initially suggested.
2. the test case duration is being reset at the beginning of each run. While that is correct from the point of view of the single test case, it may not correctly add up to the total duration of the test suite run(?)
I remember that's the case when running tests with `--parallel`.

Anyway, checks on Travis CI seem to randomly fail (?) but I can't manually re-trigger them (unless I do commit).

src/runtime/test_case_runner.js
```diff
@@ -166,26 +172,40 @@ export default class TestCaseRunner {
...
+      await this.runHooks(
+        this.afterHookDefinitions,
+        {
+          sourceLocation: this.testCaseSourceLocation,
+          pickle: this.testCase.pickle,
+          result: this.result,
+        },
+        false
+      )
+      const shouldRetryTestCase =
```
charlierudolph: Can we rename this to `shouldRetry`? I don't think the TestCase suffix is needed

src/runtime/test_case_runner.js
```diff
@@ -166,26 +172,40 @@ export default class TestCaseRunner {
...
-      },
-      false
-    )
-    this.emit('test-case-finished', { result: this.result })
+    let retries = 0
+    for (; retries <= this.retries; retries++) {
+      this.emit('test-case-started', {})
+      if (retries > 0) {
+        this.resetTestProgressData()
+        this.result.retryAttempt = retries
```
charlierudolph: I'd rather have this be a top level property on test-case-started and test-case-finished instead of in the result. Thoughts on naming it attemptNumber as well and always passing it?
```js
// Set this.maxAttempts = 1 + retries in the constructor
let attemptNumber = 1;
this.emit('test-case-started', {attemptNumber})
this.emit('test-case-finished', {attemptNumber, result: this.result})
```

src/runtime/test_case_runner.js
```diff
@@ -166,26 +172,40 @@ export default class TestCaseRunner {
...
+        },
+        false
+      )
+      const shouldRetryTestCase =
+        this.result.status === Status.FAILED && retries < this.retries
+      if (!shouldRetryTestCase) {
+        this.emit('test-case-finished', { result: this.result })
+        break
+      }
+      this.result.status = Status.FLAKY
```
charlierudolph: I'm still questioning this status as its not flaky unless a later attempt passes. Thoughts on instead adding an additional boolean to the result `retried` instead of introducing this status?

src/runtime/test_case_runner.js
```diff
@@ -166,26 +172,40 @@ export default class TestCaseRunner {
...
+      this.emit('test-case-finished', { result: this.result })
```
charlierudolph: Thoughts on rearranging this a bit and bringing the reset of the test progress data down here? 
```js
const shouldRetry = // ...
if (shouldRetry) {
  // update result status
}
// emit result
if (shouldRetry) {
  // reset progress data 
} else {
  break
}
```

src/runtime/test_case_runner.js
chipironcin: You dont know if you have retried until you retry. But we cannot mark it as failed because it will fake the whole report...
Can we mark it as "retried" only if "retries" and "number of retries" are in use?

src/runtime/test_case_runner.js
hurrikam: Done

src/runtime/test_case_runner.js
hurrikam: Done

src/runtime/test_case_runner.js
hurrikam: I've grouped all the `shouldRetry` logic at the bottom of the loop

src/runtime/test_case_runner.js
hurrikam: I've renamed `FLAKY` status into `RETRY` to be semantically more correct.

src/formatter/summary_formatter.js
```diff
@@ -14,9 +14,11 @@ export default class SummaryFormatter extends Formatter {
...
 
   isTestCaseWarning(testCase) {
-    return _.includes(
-      [Status.PENDING, Status.UNDEFINED],
-      testCase.result.status
+    return (
+      _.includes(
+        [Status.RETRY, Status.PENDING, Status.UNDEFINED],
+        testCase.result.status
+      ) || testCase.attemptNumber > 1
```
charlierudolph: Do we need to check the attempt number? Seems like that would show later passed attempts. Can we have a test for the changes here?

src/formatter/helpers/issue_helpers.js
```diff
@@ -171,3 +173,15 @@ export function formatIssue({
...
+  const result = testCase.result
+  if (!testCase.result) {
+    return ''
+  }
+  const attemptNumber = testCase.attemptNumber
+  if (attemptNumber > 1 || result.status === Status.RETRY) {
+    return flakyColorFn(`(attempt #${attemptNumber}) `)
+  }
+  return ''
+}
```
charlierudolph: Can we please add a test for this update?

src/formatter/helpers/issue_helpers.js
```diff
@@ -171,3 +173,15 @@ export function formatIssue({
...
   return `${text}\n`
 }
+
+function getRetryWarningText(testCase, flakyColorFn) {
+  const result = testCase.result
+  if (!testCase.result) {
+    return ''
+  }
+  const attemptNumber = testCase.attemptNumber
+  if (attemptNumber > 1 || result.status === Status.RETRY) {
```
charlierudolph: Again don't think we need to test the attemptNumber.

src/formatter/helpers/issue_helpers.js
```diff
@@ -171,3 +173,15 @@ export function formatIssue({
...
+    return flakyColorFn(`(attempt #${attemptNumber}) `)
```
charlierudolph: Thoughts on having this be `(attempt X, retried)`

src/formatter/progress_bar_formatter.js
```diff
@@ -38,7 +39,9 @@ export default class ProgressBarFormatter extends Formatter {
   }
 
   logErrorIfNeeded({ sourceLocation, result }) {
-    if (isIssue(result.status)) {
+    const status = result.status
+    const shouldLog = isIssue(status) || status === Status.RETRY
```
charlierudolph: Can we have isIssue account for the retry status instead of accounting for it here?

hurrikam: I didn't want to push the `RETRY` check into`isIssue`, as I guess that may be used to determine whether a test should be treated as a failure and stop execution.
If that's not the case I agree to move the check inside, even though I would still question whether retry at that point is to be considered an issue.

hurrikam:
@charlierudolph no problem about your comments. I want everyone to be happy with the implementation as well. The only time I have for now to look at the changes is more likely during weekends, so allow a few days for me to pick them. :)

src/formatter/summary_formatter.js
hurrikam: I will review this, since I added a check for the `RETRY` status, testing the attempt number may just be a left over.

src/formatter/helpers/issue_helpers.js
hurrikam: Sure

src/formatter/helpers/issue_helpers.js
hurrikam: Same as previous reply: `attemptNumber` may no longer be needed.

src/formatter/helpers/issue_helpers.js
hurrikam: Are you proposing:

`` return flakyColorFn(`(attempt ${attemptNumber}, retried) `) ``

?

chipironcin: I think he just want to specifically mention that the test has been retried, despite "attempt #x" already posing so.
Maybe we can have

    return flakyColorFn(`(test retried, attempt #${attemptNumber})`)

hurrikam: @chipironcin your code provides the same information as mine, just inverting the order ;)

hurrikam:
Hi @charlierudolph, I've finally had the time to get back to the PR.
I had a look back at the Ruby implementation and I decided to remove the `RETRY` status and re-introduce the `FLAKY` one:

https://www.rubydoc.info/github/cucumber/cucumber-ruby-core/Cucumber/Core/Test/Result/Flaky

While flaky is actually a status, retry is just a temporary flag to indicate the test should-rerun and I don't think it was needed in there.

I've also reviewed the logic of `test_case_runner` to include `attemptNumber` when triggering events ONLY if the test case accounts for retries. The reason is:
1. The attempt number information is redundant when a test is meant to run just once (no retry, default behaviour)
2. Having the attempt number conditionally emitted allows formatters to print the retry information only when meaningful

As per the formatters:

1. I have added more unit test coverage around events protocol
2. The `Summary` formatter will print (similary to Ruby): `2 scenarios (1 flaky, 1 passed)`
3. The `Progress-Bar` formatter will warn about each failing run of a flaky test.
4. I've simplified the conditional warning following a scenario name to just say `(attempt X)`. If we want to keep it in the format  `(attempt X, retried)` then we need a way to check whether the run is not the first one or the last one, depending of what that 'retried' means. However, I think just printing the attempt number is all the user needs to know.

Please let me know if you're fine with the current implementation so we can work on remaining details or test cases, if anything is missing.

features/retry.feature
```diff
@@ -0,0 +1,103 @@
...
+      Given(/^a flaky step$/, function() {
+        if (willPass) {
+          return
+        }
+        willPass = true
+        throw 'fail'        
+      })
+      """
+    When I run cucumber-js with `--retry 1`
+    Then the step "a flaky step" has status "passed"
```
charlierudolph: I think we should test that this step / scenario ran twice one failing, one passing. Same for below test that we ran two scenarios and both failed.

features/retry.feature
```diff
@@ -0,0 +1,103 @@
+Feature: Retry flaky tests
+
+  Using the `--retry` flag will retry failing tests for the specified number of times
+  Additionally using the `--retryTagFilter` flag will re-run only tests matching the tag expression
+
+  @spawn
+  Scenario: running Cucumber JS with --retryTagFilter but no positive --retry will fail
+    When I run cucumber-js with `--retryTagFilter @flaky`
+    Then it fails
```
charlierudolph: I think we should test here that it gives useful output.

src/formatter/progress_bar_formatter.js
```diff
@@ -38,7 +38,8 @@ export default class ProgressBarFormatter extends Formatter {
   }
 
   logErrorIfNeeded({ sourceLocation, result }) {
-    if (isIssue(result.status)) {
+    const shouldLog = isIssue(result.status)
```
charlierudolph: I don't think the change to this file is needed anymore

src/formatter/helpers/issue_helpers.js
```diff
@@ -171,3 +174,11 @@ export function formatIssue({
   })
   return `${text}\n`
 }
+
+function getRetryWarningText(testCase, flakyColorFn) {
+  const attemptNumber = testCase.attemptNumber
```
charlierudolph: This can be
`const { attemptNumber } = testCase`

src/formatter/helpers/issue_helpers.js
charlierudolph: Yep

hurrikam:
Hi @charlierudolph thanks for the feedback. I tried to follow the Ruby implementation where the `flaky` flag is used.
Assuming we mark a failing test run as `retried` if it will be retried, then what about the final one?
Then that should be either `failed` or `passed` as that won't be `retried`. We have the `attemptNumber` to tell us if and how many times a test has been retried.

charlierudolph:
> Assuming we mark a failing test run as `retried` if it will be retried, then what about the final one?
Then that should be either `failed` or `passed` as that won't be retried. We have the `attemptNumber` to tell us if and how many times a test has been retried.

Yes the final one would be `failed` or `passed` and the final one is the only thing that is used to determine how that test case effects the overall test run. The attempt number will only be used by the formatters when outputting the retried warnings / final failure.

hurrikam:
Hi guys, apologies but I've been really busy recently. I've been working on the latest PR changes on a different branch I'm planning to merge once I double check all tests are passing. Please bear with me :)

Javrd:
Using this branch I got this warning again:
> (node:14214) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

I had also it using 3.0.4 cucumber version and it stops when I updated my project to 5.1.0 version.

hurrikam:
@Javrd the branch is quite old and it needs to be update from master. On top of that, the feature is still under refactoring :)

chipironcin:
Useful "working" fork:
https://github.com/Nick-Lucas/cucumber-js.git#feature/issue-727-retry

Nick-Lucas:
Just a heads up @chipironcin I've now left the company which I forked that for, and transferred the repo to their github: https://github.com/wonderbill/cucumber-js

My fork will disappear soon, so either fork it yourself, or use the new fork. :)

hurrikam:
Hi @chipironcin and @Nick-Lucas. I've been really busy at work and with personal projects in the last months and eventually I forgot about this PR :)
Are you still planning to use this branch or should we forget it?

samtbabu:
Waiting to get this Released!

hurrikam:
Nice! 😄

chipironcin:
I am still interested too. Seems there is a failing test in AppVeyor.
Also might be worth to take a look at the ?working? fork in https://github.com/wonderbill/cucumber-js in case we are missing something

hurrikam:
@chipironcin I've checked https://github.com/wonderbill/cucumber-js/commits/feature/issue-727-retry and the only two commits on top of my last changes seem to contain only built/webpacked code. Have a look, just to make sure I'm not missing anything.

Nick-Lucas:
Yeah we literally just removed the gitignore and committed the build files so the branch could just be checked out in package.json. it's also quite old code now, a bunch of refactoring took place on this end since.

hurrikam:
@samtbabu when is this going to be released? :)

Znakki:
@samtbabu do you have any news about releasing this feature?

individual-it:
I've rebased, fixed conflicts and fixed CI for this PR in https://github.com/cucumber/cucumber-js/pull/1205
please review and merge if possible

aslakhellesoy:
Superseded by #1205

PR #1123, opened 2018-08-30
Resolve some licensing problems in dependencies

honzajavorek:
Follow up to https://github.com/cucumber/cucumber-js/issues/1124. See each commit message for a detailed reasoning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,7 +142,7 @@
     "assertion-error-formatter": "^2.0.1",
     "babel-runtime": "^6.11.6",
     "bluebird": "^3.4.1",
-    "cli-table": "^0.3.1",
+    "cli-table2": "^0.2.0",
     "colors": "^1.1.2",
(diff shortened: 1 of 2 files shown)


honzajavorek:
I don't understand the dependency lint report. I'm happy to fix the PR, but I'm not sure what I'm supposed to do.

charlierudolph:
Can we separate this into two PRs? One that updates cli-table and one that updates string-argv? 

I'm good with the string-argv updates.

For the cli-table updates, why are we changing the dependency? If we are, we need to update the source code to use the new dep. Dependency lint is reporting that cli-table is required in the source code but is not installed as a dependency and that cli-table2 is installed but not used.

honzajavorek:
Ah! I get it now. I forgot to update the code. I’ll do that, thanks. I
explain the changes in the issue I mentioned.

PR #1126, opened 2018-09-03
Replace cli-table with cli-table2

honzajavorek:
The cli-table looks like an abandoned project and has licensing issues it is not willing to resolve:

- https://github.com/Automattic/cli-table/issues/93
- https://github.com/Automattic/cli-table/pull/101
- https://github.com/Automattic/cli-table/pull/68

According to https://github.com/Automattic/cli-table/pull/68#issuecomment-268793354 it is advised to use [cli-table2](https://github.com/jamestalmage/cli-table2/) instead. Moreover, that package has licensing all right.

Related to https://github.com/cucumber/cucumber-js/issues/1124

PR #1131, opened 2018-09-11
Replace cli-table2 with cli-table3

honzajavorek:
Addressing https://github.com/cucumber/cucumber-js/pull/1126#issuecomment-419610002


honzajavorek:
Thank you!

PR #1173, opened 2019-02-07
Add cucumber expressions to step definitions docs.

### LINKED ISSUE #1172: Documentation: add examples with the string syntax for defining steps
Currently the doc for steps definition only shows the RegExp syntax for defining steps https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/step_definitions.md

Expected: Both the RegExp and the string syntax should be shown, a link to a complete explanation of the string syntax should be available.

I can make a PR adding an example with string syntax but I'm not sure where all the possible template types are defined (I just learned about `{word}`)

Izhaki:
Resolves #1172

### DIFF
--- a/docs/support_files/step_definitions.md
+++ b/docs/support_files/step_definitions.md
@@ -1,7 +1,32 @@
 # Step Definitions
 
-Step definitions are glue between features written in Gherkin and the actual system under test.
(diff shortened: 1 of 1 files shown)


Izhaki:
@charlierudolph

PR #1273, opened 2019-12-30
format: add support for gherkin rule/example syntax

davidjgoss:
After the move to Gherkin 9 and cucumber-messages, we can execute specs that include Rule/Example syntax, but still need to update the formatter helpers and built-in formatters to handle the results.


features/fixtures/formatters/passed-rule.message.json.ts
```diff
@@ -0,0 +1,142 @@
...
+        ],
+      },
+    },
+  },
+  {
+    pickle: {
+      id: '4',
+      uri: 'features/a.feature',
+      name: 'an example',
+      language: 'en',
```
davidjgoss: Should the pickle name here be some kind of concatenation of the rule name and example name, or is the example name sufficient?

features/fixtures/formatters/passed-rule.json.ts
```diff
@@ -0,0 +1,37 @@
+module.exports = [
+  {
+    description: '',
+    elements: [
```
davidjgoss: Given the JSON formatter’s status as soon-to-be-deprecated, and other tools’ reliance on it, would seem best to not add any additional structure for Rule/Example, and instead just fit within the roaring structure for scenarios - do we think that’s right?

davidjgoss:
The happy path for JSON formatter is working, moving on to failure cases and other built-in formatters, a couple of bits I am unsure of (above) which can be circled back to.

davidjgoss:
@charlierudolph this is ready to go now I think.

In the end it was mostly about changing gherkin document helper to be able to handle the rule/example structure when matching up from pickles, and some usage of this in the JSON formatter - beyond that it's just adding/changing tests on the other formatters to confirm they work normally, and adding some acceptance-level coverage as well.

Let me know if you think anything is missing.

davidjgoss:
@charlierudolph this will probably need a rebase after #1286 goes in, but otherwise is it okay?

src/formatter/progress_formatter_spec.ts
```diff
@@ -31,10 +31,11 @@ Feature: a
...
     Given a passing step
-  Scenario: a5
-    Given a skipped step
-  Scenario: a6
-    Given an undefined step
+  Rule: ab
+    Example: a5
+      Given a skipped step
+    Example: a6
+      Given an undefined step
```
charlierudolph: Can we please make a new test for this vs combining into one? This was really for each step status and I think its useful to have rule as a separate test

src/formatter/helpers/gherkin_document_parser_spec.ts
```diff
@@ -0,0 +1,1294 @@
...
+  getGherkinScenarioMap,
+  getGherkinStepMap,
+} from './gherkin_document_parser'
+
+describe('GherkinDocumentParser', () => {
+  describe('with a Background and Scenario', () => {
+    let sourceData
+    beforeEach(() => {
+      // Arrange
+      sourceData = {
```
charlierudolph: Really don't like this as we are hardcoding some of gherkin's internal structure in a test that would need to be updated if gherkin changes something. 

Instead of hardcoding it, can we generate this by having the gherkin library parse some content?

Also when asserting the output, can we have maybe only one test assert the full output once to show structure and then later tests can assume the structure is preserved and perform simpler assertions?

Lastly, for general test structure I'm leaning towards no longer using beforeEach and instead just making helper functions so everything you need to know about the test is in the block (instead of having to find beforeEach blocks that setup some state). Thoughts on that?

features/fixtures/formatters/failed-rule.json.ts
```diff
@@ -0,0 +1,39 @@
+module.exports = [
```
charlierudolph: Do we really need these? I only created json fixtures to make it easy to see if we changed something for backward compatibility. Maybe just one for a rule would be enough?

src/formatter/helpers/gherkin_document_parser_spec.ts
davidjgoss: > Instead of hardcoding it, can we generate this by having the gherkin library parse some content?

Yeah that's much better, will do

> can we have maybe only one test assert the full output once to show structure

Makes sense, will try

> leaning towards no longer using beforeEach and instead just making helper functions

Yeah I see where you're coming from there. I think it would probably improve it in this case. I have seen the `beforeEach` pattern work well where it's use for arrange and act, and then the `it`s that follow it are just a line or two each of asserts.

src/formatter/progress_formatter_spec.ts
davidjgoss: Done

features/fixtures/formatters/failed-rule.json.ts
davidjgoss: Removed so we are now only adding one json fixture for rule happy path

src/formatter/helpers/gherkin_document_parser_spec.ts
davidjgoss: Okay, I've reworked this now based on the above. So we now parse each example feature with gherkin, and then assert on its output with references - so whilst we are still somewhat coupled to the hierarchical structure of a gherkin document, it will be an obvious compilation error if that changes, rather than assertions being out. Let me know what you think.

davidjgoss:
@charlierudolph feedback addressed and latest changes from master merged in here

features/fixtures/formatters/failed-rule.json.ts
charlierudolph: Still seeing failed rule and rejected pickle rule

src/formatter/json_formatter.ts
```diff
@@ -5,15 +5,20 @@ import { formatLocation, GherkinDocumentParser, PickleParser } from './helpers'
...
 import { messages } from 'cucumber-messages'
-import { getGherkinScenarioLocationMap } from './helpers/gherkin_document_parser'
+import {
+  getGherkinExampleRuleMap,
+  getGherkinScenarioLocationMap,
+} from './helpers/gherkin_document_parser'
 import { ITestCaseAttempt } from './helpers/event_data_collector'
 import { doesHaveValue, doesNotHaveValue } from '../value_checker'
 import { parseStepArgument } from '../step_arguments'
+import { io } from 'cucumber-messages/dist/src/cucumber-messages'
```
charlierudolph: Please use `import {messages} from 'cucumber-messages'`

src/formatter/json_formatter.ts
```diff
@@ -70,6 +75,9 @@ interface IBuildJsonFeatureOptions {
 interface IBuildJsonScenarioOptions {
   feature: messages.GherkinDocument.IFeature
   gherkinScenarioMap: Dictionary<messages.GherkinDocument.Feature.IScenario>
+  gherkinExampleRuleMap: Dictionary<
+    messages.GherkinDocument.Feature.FeatureChild.IRule
```
charlierudolph: Could this be simplified to `IRule`?

src/formatter/progress_bar_formatter_spec.ts
```diff
@@ -94,7 +94,7 @@ describe('ProgressBarFormatter', () => {
         },
         {
           data:
-            'Feature: a\nScenario: b\nGiven a step\nWhen a step\nThen a step',
+            'Feature: a\nRule: b\nExample: c\nGiven a step\nWhen a step\nThen a step',
```
charlierudolph: Can we please create an additional test for this? (and retain the old one)

src/formatter/rerun_formatter_spec.ts
```diff
@@ -180,7 +180,7 @@ describe('RerunFormatter', () => {
                 uri: 'a.feature',
               },
               {
-                data: 'Feature: a\n\nScenario: b\nGiven a step',
+                data: 'Feature: a\n\nRule: b\nExample: c\nGiven a step',
```
charlierudolph: Same, please add new test

src/formatter/helpers/gherkin_document_parser_spec.ts
charlierudolph: Looks good. Thanks for the update

features/fixtures/formatters/failed-rule.json.ts
davidjgoss: Ah okay - I think I thought you were only talking about the fixtures for the JSON formatter. Should I scale back the ones for the message formatter as well?

src/formatter/json_formatter.ts
davidjgoss: Done

src/formatter/json_formatter.ts
davidjgoss: Yep, done

features/fixtures/formatters/failed-rule.json.ts
charlierudolph: Haha understandable. I meant the fixtures in general (both json + message formatter). Think just a passed version of json / messages is sufficient

src/formatter/progress_bar_formatter_spec.ts
davidjgoss: Done

src/formatter/rerun_formatter_spec.ts
davidjgoss: Done

features/fixtures/formatters/failed-rule.json.ts
davidjgoss: No worries, done that now.

features/formatters.feature
```diff
@@ -1,17 +1,17 @@
...
+  Scenario: rejected pickle from Scenario
     Given a file named "features/a.feature" with:
       """
       Feature: a feature
         Scenario: a scenario
           Given a step
       """
     When I run cucumber-js with all formatters and `--tags @a`
-    Then the "message" formatter output matches the fixture "formatters/rejected-pickle.message.json"
+    Then the "message" formatter output matches the fixture "formatters/rejected-pickle-scenario.message.json"
```
charlierudolph: can we also revert the changes to this scenario and the failed scenario so that we only specify the difference for scenario vs rule for the passed scenarios?

src/formatter/progress_formatter_spec.ts
```diff
@@ -31,7 +31,7 @@ Feature: a
     Given a pending step
   Scenario: a4
     Given a passing step
-  Scenario: a5
+ Scenario: a5
```
charlierudolph: please revert the whitespace change here

PR #1286, opened 2020-02-03
build: add noImplicitAny and noImplicitReturns, fix compilation

davidjgoss:
Part of #1269 

`noImplicitReturns` ended up being no changes. `noImplicitAny` on the other hand needed quite a lot.

- - - 

- [x] Figure out why `fs-extra` typings don't line up with code
- [x] Figure out why `unit-test` fails on not picking up our external module declarations


features/step_definitions/file_steps.ts
```diff
@@ -7,37 +7,46 @@ import path from 'path'
 import Mustache from 'mustache'
 import { World } from '../support/world'
 
-Given(/^a file named "(.*)" with:$/, function(
+Given(/^a file named "(.*)" with:$/, async function(
   this: World,
-  filePath,
-  fileContent
+  filePath: string,
+  fileContent: any
```
charlierudolph: Shouldn't fileContent be a string?

features/support/hooks.ts
```diff
@@ -37,6 +37,7 @@ Before(function(
 
   const tmpDirNodeModulesPath = path.join(this.tmpDir, 'node_modules')
   const tmpDirCucumberPath = path.join(tmpDirNodeModulesPath, 'cucumber')
+  // @ts-ignore
```
charlierudolph: What is being ignored here?

src/cli/i18n.ts
```diff
@@ -14,7 +14,7 @@ const keywords = [
   'then',
   'and',
   'but',
-]
+] as const
```
charlierudolph: What makes this needed?

src/formatter/json_formatter.ts
```diff
@@ -9,6 +9,11 @@ import { getGherkinScenarioLocationMap } from './helpers/gherkin_document_parser
 import { ITestCaseAttempt } from './helpers/event_data_collector'
 import { doesHaveValue, doesNotHaveValue } from '../value_checker'
 import { parseStepArgument } from '../step_arguments'
+import ITag = messages.GherkinDocument.Feature.ITag
+import IFeature = messages.GherkinDocument.IFeature
+import IPickle = messages.IPickle
+import IScenario = messages.GherkinDocument.Feature.IScenario
+import IEnvelope = messages.IEnvelope
```
charlierudolph: Hmm, why are these import vs const?

src/cli/i18n.ts
davidjgoss: Later in the file, we iterate over these keywords and get the value of each from a gherkin dialect object:

```ts
const language = Gherkin.dialects()[isoCode]
const rows = _.map(keywords, keyword => {
  const words = _.map(language[keyword], s => `"${s}"`).join(', ')
  return [capitalCase(keyword), words]
})
```

For the TS compiler to be okay with that, we have to add the `as const` here to tell it that the array won't be mutated, making it type safe because all of these keywords are properties of a dialect with string values.

src/formatter/json_formatter.ts
davidjgoss: Yeah, I think you're right, I'll change and see if there's an eslint rule for that

features/support/hooks.ts
davidjgoss: `fs-extra` typings seem to be out of sync with the code we are using - this is one I'm looping back to once I get the rest it green

features/step_definitions/file_steps.ts
davidjgoss: Yep it should, will fix

src/formatter/json_formatter.ts
davidjgoss: Okay this didn't actually work, so kept as it is. I'm trying to figure out why, maybe something relating to namespaces.

features/support/hooks.ts
davidjgoss: This is a strange one - the `createSymlink` and `createSymlinkSync` methods have been on fs-extra [for years](https://github.com/jprichardson/node-fs-extra/blame/master/lib/ensure/index.js) but they aren't documented and aren't in the DefinitelyTyped definitions.

I'll make PRs for both, link them up here, and see what comes out of it.

src/cli/i18n.ts
charlierudolph: But the array is already saved into a const variable. Would providing a type to that like
```typescript
const keywords: string[] = [ /* ... */ ]
```
work?

davidjgoss: That doesn't quite do it because although declaring with `const` means it can never be reassigned, the array contents can still be modified in place:

image

The `as const` thing types it as an immutable array:

image

This means the compiler is satisfied that it's safe to use the elements as keys of a dialect.

features/support/hooks.ts
davidjgoss: I've switched our code to use `createSymlinkSync` which has the same outcome, and raised https://github.com/jprichardson/node-fs-extra/issues/758 to query the API status of the other method.

features/step_definitions/fixture_steps.ts
```diff
@@ -6,13 +6,15 @@ import {
 } from '../support/formatter_output_helpers'
 import fs from 'mz/fs'
 import path from 'path'
+import { messages } from 'cucumber-messages'
+import Envelope = messages.Envelope
```
charlierudolph: I'm a bit torn on this. I understand the idea of using this to dry up code usage but also don't like the idea of having a ton of imports up here as we list all the imports we are using (I think one file has at least 5 of these).

I think I'd favor us not having this and instead always using `messages.*` in the code. I think that helps make it easier to tell where the interface comes from (in our code vs a library). Not every library deals with its interfaces in this way so that idea doesn't really apply in other places.

Fine to leave as is for this PR, but I'll maybe create an issue for us to try to come to an agreement on if we want to do this going forward or not and migrate code to whichever pattern we choose

features/support/world.ts
```diff
@@ -63,7 +63,7 @@ export class World {
       const cli = new Cli({
         argv: args,
         cwd,
-        stdout,
+        stdout: (stdout as unknown) as WriteStream,
```
charlierudolph: Why is this needed? Is stdout not a WriteStream?

src/cli/i18n.ts
charlierudolph: Hmm okay. Don't think that is something we really need and dislike the uncommon syntax but I'm okay leaving it.

src/formatter/get_color_fns.ts
```diff
@@ -23,6 +23,7 @@ export default function getColorFns(enabled: boolean): IColorFns {
           [Status.PENDING]: colors.yellow.bind(colors),
           [Status.SKIPPED]: colors.cyan.bind(colors),
           [Status.UNDEFINED]: colors.yellow.bind(colors),
+          [Status.UNKNOWN]: colors.yellow.bind(colors),
```
charlierudolph: When is this used?

src/formatter/helpers/event_data_collector.ts
```diff
@@ -20,13 +21,18 @@ export interface ITestCaseAttempt {
   testCase: messages.ITestCase
 }
 
+export interface IEventBroadcaster {
+  on(name: string, handler: Function): void
+  emit(name: string, envelope: IEnvelope): void
+}
```
charlierudolph: Don't think this is needed. Have just used `EventEmitter` thus far. Can we stick with that for now and if we want to change that do so in a separate PR?

src/formatter/helpers/keyword_type.ts
```diff
@@ -1,6 +1,7 @@
 import _ from 'lodash'
 import Gherkin from 'gherkin'
 import { doesHaveValue } from '../../value_checker'
+import Dialect from 'gherkin/dist/src/Dialect'
```
charlierudolph: Is there no other way to access this without reaching into Gherkin? If not, lets open an issue in cucumber if that doesn't exist already

src/models/test_case_hook_definition.ts
```diff
@@ -3,18 +3,22 @@ import Definition, {
   IDefinition,
   IGetInvocationDataResponse,
   IGetInvocationDataRequest,
+  IDefinitionParameters,
+  IHookDefinitionOptions,
 } from './definition'
+import { io } from 'cucumber-messages/dist/src/cucumber-messages'
```
charlierudolph: Please use `import { messages } from 'cucumber-messages'`

src/models/test_case_hook_definition_spec.ts
```diff
@@ -10,6 +10,11 @@ describe('TestCaseHookDefinition', () => {
         // Arrange
         const pickle = await getPickleWithTags([])
         const testCaseHookDefinition = new TestCaseHookDefinition({
+          code: undefined,
+          id: '',
+          line: 0,
+          uri: '',
+
```
charlierudolph: please remove this blank line. Appears a few times in this file

src/pickle_filter.ts
```diff
@@ -4,6 +4,9 @@ import parse from 'cucumber-tag-expressions'
 import { getGherkinScenarioLocationMap } from './formatter/helpers/gherkin_document_parser'
 import { messages } from 'cucumber-messages'
 import { doesHaveValue, doesNotHaveValue } from './value_checker'
+import { io } from 'cucumber-messages/dist/src/cucumber-messages'
```
charlierudolph: Please use `import { messages } from 'cucumber-messages'`

src/runtime/helpers_spec.ts
```diff
@@ -10,12 +10,21 @@ describe('Helpers', () => {
     beforeEach(function() {
       this.result = getAmbiguousStepException([
         new StepDefinition({
-          line: '3',
+          code: undefined,
+          expression: undefined,
+          id: '',
+          options: undefined,
+
```
charlierudolph: please remove this blank line

src/runtime/pickle_runner_spec.ts
```diff
@@ -7,11 +7,13 @@ import { messages } from 'cucumber-messages'
 import { incrementing } from 'cucumber-messages/dist/src/IdGenerator'
 import { parse } from '../../test/gherkin_helpers'
 import { buildSupportCodeLibrary } from '../../test/runtime_helpers'
-import lolex from 'lolex'
+import lolex, { InstalledClock } from 'lolex'
 import timeMethods, { millisecondsToDuration, getZeroDuration } from '../time'
 import { getBaseSupportCodeLibrary } from '../../test/fixtures/steps'
 import { ISupportCodeLibrary } from '../support_code_library_builder/types'
 import { valueOrDefault } from '../value_checker'
+import { io } from 'cucumber-messages/dist/src/cucumber-messages'
```
charlierudolph: Please use `import { messages } from 'cucumber-messages'`

src/support_code_library_builder/validate_arguments.ts
```diff
@@ -1,30 +1,37 @@
...
+interface IDefineStepArguments {
+  pattern?: DefineStepPattern
+  options?: IDefineStepOptions
+  code?: Function
+}
+
 const optionsValidation = {
   expectedType: 'object or function',
-  predicate({ options }) {
+  predicate({ options }: Partial<IDefineStepArguments>) {
```
charlierudolph: I don't think we need `Partial` given all the fields are optional

src/support_code_library_builder/world.ts
```diff
@@ -1,10 +1,17 @@
 import { IAttachment } from '../runtime/attachment_manager'
 
+export type IAttachFunction = (attachment: IAttachment) => void
```
charlierudolph: This should probably live in `src/runtime/attachment_manager/index.ts`

test/formatter_helpers.ts
```diff
@@ -10,6 +10,8 @@ import { ITestCaseAttempt } from '../src/formatter/helpers/event_data_collector'
 import { doesNotHaveValue } from '../src/value_checker'
 import { IParsedArgvFormatOptions } from '../src/cli/argv_parser'
 import { PassThrough } from 'stream'
+import { io } from 'cucumber-messages/dist/src/cucumber-messages'
```
charlierudolph: Please use `import { messages } from 'cucumber-messages'`

test/gherkin_helpers.ts
```diff
@@ -15,6 +16,9 @@ export interface IParsedSourceWithEnvelopes extends IParsedSource {
 export async function parse({
   data,
   uri,
+}: {
+  data: string
+  uri: string
```
charlierudolph: Can we create an `IParseRequest`? Same for below

features/step_definitions/fixture_steps.ts
davidjgoss: Yeah that makes sense. We could prolly use `eslint-plugin-import` to enforce whatever we settle on too.

src/formatter/get_color_fns.ts
davidjgoss: Error with the `UNKNOWN` one removed:

```
src/formatter/get_color_fns.ts:19:16 - error TS7053: Element implicitly has an 'any' type because expression of type 'Status' can't be used to index type '{ [Status.AMBIGUOUS]: any; [Status.FAILED]: any; [Status.PASSED]: any; [Status.PENDING]: any; [Status.SKIPPED]: any; [Status.UNDEFINED]: any; }'.
  Property '[Status.UNKNOWN]' does not exist on type '{ [Status.AMBIGUOUS]: any; [Status.FAILED]: any; [Status.PASSED]: any; [Status.PENDING]: any; [Status.SKIPPED]: any; [Status.UNDEFINED]: any; }'.
```

I'm not sure if this function is ever called with `UNKNOWN`, but the compiler's point is that it _could_ happen at runtime, and since the function signature requires it to return an `IColorFn`, we have to cater to all values in the enum in our object.

(I keep running into smarts I didn't know TypeScript had here...)

src/models/test_case_hook_definition.ts
davidjgoss: Done

src/pickle_filter.ts
davidjgoss: Done

src/runtime/pickle_runner_spec.ts
davidjgoss: Done

test/formatter_helpers.ts
davidjgoss: Done

src/runtime/helpers_spec.ts
davidjgoss: Done

src/models/test_case_hook_definition_spec.ts
davidjgoss: Done

src/support_code_library_builder/validate_arguments.ts
davidjgoss: Yep, fixed

src/support_code_library_builder/world.ts
davidjgoss: Sure, moved

test/gherkin_helpers.ts
davidjgoss: Done and done

features/support/world.ts
davidjgoss: We're using a `PassThrough` here, which doesn't extend either the fs or tty `WriteStream` although it has all the methods we use.  Unfortunately the closest shared interface between all three is `Writable` but that that is missing stuff and would give us compilation issues.

We can avoid this by adding `PassThrough` to the `IFormatterStream` interface - I have done that for now, let me know what you think.

davidjgoss:
@charlierudolph all addressed

src/formatter/helpers/keyword_type.ts
charlierudolph: Cool. Thanks for creating an upstream PR!

features/support/world.ts
```diff
@@ -63,7 +63,7 @@ export class World {
       const cli = new Cli({
         argv: args,
         cwd,
-        stdout,
+        stdout: stdout,
```
charlierudolph: This doesn't seem to be needed

src/runtime/attachment_manager/index.ts
```diff
@@ -8,6 +8,8 @@ export interface IAttachment {
   media: messages.Media
 }
 
+export type IAttachFunction = (attachment: IAttachment) => void
```
charlierudolph: We can now use this to DRY up the `onAttachment` property  / attachment manager constructor

davidjgoss: Yup done

Issue #656, opened 2016-10-20
Timeout and Promise causing "Uncaught Error" in console


Sawtaytoes:
I was unable to reproduce this issue in a command line project, but I was able to using the website HTML example: http://cucumber.github.io/cucumber-js/

I only changed the `When` part in the sample code, but it happens on `Given` and `Then` as well:

``` js
this.When(/^I increment the variable by (\d+)$/, function(number, next) {
  this.incrementBy(number);
  setTimeout(() => {
    throw 'this is an uncaught error'
    next()
  }, 1000)
});
```

Looking at the console when running the feature, I get the message:

```
Uncaught this is an uncaught error
```

It never goes to `next()`, so it take 5 seconds to finish. The same issue happens in `Promise.then` and also in `setInterval`. I also tested when returning a `Promise` and not passing `next`, but experienced the same issue.

I use Chai as my assertion library, and when I do `expect(true).to.be.false`, it throws and `Assertion Error` and normally; that gets picked up through Cucumber which properly fails the test. On the other hand, if I wrap this error in a `setTimeout`, `setInterval`, or `Promise`, then the `Uncaught` error message shows up in the console instead of Cucumber picking up on the error.

davidrissato:
Sawtaytoes,

The way you did it will never work for sure and cucumber it is behaving exactly as expected.

When dealing with asynchronous code the way you did (using callbacks), you need to try...catch exceptions and pass it to next method, like that:

```javascript
this.When(/^I increment the variable by (\d+)$/, function(number, next) {
  this.incrementBy(number);
  setTimeout(() => {
    try {
        throw new Error('this is an uncaught error');
        next(); // just for the example, of course it will never be called
    } catch (err) {
       next(err);
    }
  }, 1000)
});
```

However, there are much better ways to deal with asynchronous code in Javascript. I strongly suggest you invest some time researching  "javascript promises" and also take a look in bluebird.js package (a nice Promise implementation). It will well worth your time investment and will strongly leverage your code quality.

Best

Sawtaytoes:
That's the solution I ended up with. Seems sub-optimal because I have to try-catch it. I'm wondering if there's a way to extend Chai to handle this use case internally.

charlierudolph:
I personally prefer promises, but this should work the same in the CLI versus in the browser. I believe this is just a bug in the browser version.

davidrissato:
Yes, you are right. This solution is suboptimal.

[...]

PR #1318, opened 2020-05-09
update gherkin etc, validate against cck

### LINKED ISSUE #1310: Emit a Meta message at the beginning of the run.
See https://github.com/cucumber/cucumber/pull/976 for details.

It's been implemented in cucumber-ruby and cucumber-jvm:
* https://github.com/cucumber/cucumber-ruby/pull/1409
* https://github.com/cucumber/cucumber-jvm/pull/1952

davidjgoss:
More or less copied the approach from [cucumber-jvm](https://github.com/cucumber/cucumber-jvm/tree/master/compatibility).

To run:

```
$ yarn
$ yarn cck-test
```

(Or, run `compatibility/cck_spec.ts` from your IDE.)

Done:

- Expose `World` and `IWorldOptions` in public API
- Add a `log` function to the World which attaches text with the cucumber log media type
- Emit attachments in new structure
- Add some methods to `DataTable`
- Emit `Meta` message at start of run - fixes #1310 
- Keep a stopwatch of the test run, use it for a `timestamp` on various messages
- Emit messages for hooks and step definitions
- Emit messages for parameter type definitions
- `stepMatchArgumentsLists` are now being emitted within `testCase.testSteps`
- `pickleAccepted` and `pickleRejected` messages are now not emitted
- `testCaseFinished.testResult` is now not emitted (see https://github.com/cucumber/cucumber/issues/902)
- `testRunFinished.success` is now not emitted

Questions:

- `testCase` messages per CCK fixtures should be emitted together at start of run, rather than before each test case is started (non-trivial refactor - do we need this right away?)


charlierudolph:
@davidjgoss whats the state of this? Anything I can help with?

davidjgoss:
@charlierudolph almost there, just going back over some TODOs where I left things a bit unfinished around parallel runtime and retry. Hoping to finish that tonight all being well and then ready for review

davidjgoss:
@charlierudolph there are a couple of bits still to sort out here (see description) but I think this is in a good enough state to get some review especially on the new stuff added e.g. the support code messages and the stopwatch. Also the retry stuff feels a bit precarious (with the result block now no longer on a `testCaseFinished`) but it's well tested and functioning properly.

Once I get this green it's probably going to be best to pull in the latest everything again as some new releases have happened in the monorepo

features/attachments.feature
```diff
@@ -26,7 +26,7 @@ Feature: Attachments
     When I run cucumber-js
     Then scenario "some scenario" "Before" hook has the attachments:
       | DATA     | MEDIA TYPE | MEDIA ENCODING |
-      | iVBORw== | image/png  | BASE64         |
+      | iVBORw== | image/png  | 1              |
```
charlierudolph, line 29: I'd like these left as BASE64 as that is better than 1/0. What prompted this change?

src/runtime/pickle_runner_spec.ts
```diff
@@ -302,33 +329,60 @@ describe('PickleRunner', () => {
...
               id: '2',
               testCaseId: '0',
+              timestamp: {
+                nanos: 1000000,
+                seconds: {
+                  high: 0,
+                  low: 0,
+                  unsigned: false,
+                },
+              },
```
charlierudolph, line 339: Probably worth creating a test helper `getTestTimestamp` that takes in the number of milliseconds to reduce the duplication

src/runtime/stopwatch.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  stop(): ITestRunStopwatch {
+    this.stopwatch.stop()
+    return this
+  }
+
+  duration(): Duration {
+    const current = this.stopwatch.duration()
+    if (this.base !== null) {
+      return duration(this.base.nanos() + current.nanos())
```
charlierudolph, line 34: I'm confused by this. What exactly is happening here? Could we add some tests for this? (maybe allow stopwatch to be injected if mocking it out is hard)

features/attachments.feature
davidjgoss: This was because of the [message structure changing](https://github.com/cucumber/cucumber/blob/master/messages/messages.proto#L366) - thinking about it, I can probably keep a map in the step code from a string to the TypeScript enum.

src/runtime/stopwatch.ts
davidjgoss: Good point will add some tests.

The `base` thing here is used in the parallel runtime. We have a stopwatch at the test run level, but we can't pass it from the coordinator to the workers as it's not serializable, so we instead pass the current elapsed duration in nanos and then seed it to the `base` on the stopwatch that we init in the worker.

features/step_definitions/cli_steps.ts
```diff
@@ -1,12 +1,12 @@
-import { When, Then, DataTable } from '../../'
+import { DataTable, Then, When } from '../../'
```
charlierudolph, line 1: Did you reorder these manually? If we want them ordered we should consider an eslint rule. Personally don't feel strongly about it though

davidjgoss: No - I’m not sure whether that was webstorm or prettier to be honest.

features/step_definitions/message_steps.ts
```diff
@@ -23,10 +23,15 @@ type StringifiedStatus =
   | 'AMBIGUOUS'
   | 'FAILED'
 
-const { Status } = messages.TestResult
+const { Status } = messages.TestStepFinished.TestStepResult
+
+const ENCODING_MAP: { [key: string]: messages.Attachment.ContentEncoding } = {
+  IDENTITY: messages.Attachment.ContentEncoding.IDENTITY,
+  BASE64: messages.Attachment.ContentEncoding.BASE64,
+}
```
charlierudolph, line 31: I don't think this is needed. I think we could replace its usage with:

```typescript
ENCODING_MAP[x['MEDIA ENCODING']]
// becomes
messages.Attachment.ContentEncoding[x['MEDIA ENCODING']]
```

davidjgoss: Will try but don’t think so - that enum will compile down to just the ordinals, not the strings

src/models/data_table.ts
```diff
@@ -1,13 +1,19 @@
...
   private readonly rawTable: string[][]
 
-  constructor(pickleTable: messages.PickleStepArgument.IPickleTable) {
-    this.rawTable = pickleTable.rows.map((row) =>
-      row.cells.map((cell) => cell.value)
-    )
+  constructor(
+    sourceTable: messages.PickleStepArgument.IPickleTable | string[][]
+  ) {
+    if (sourceTable instanceof Array) {
```
charlierudolph, line 10: What is this condition for?

src/formatter/step_definition_snippet_builder/snippet_syntax.ts
```diff
@@ -11,7 +11,7 @@ export enum SnippetInterface {
 export interface ISnippetSyntaxBuildOptions {
   comment: string
   functionName: string
-  generatedExpressions: GeneratedExpression[]
+  generatedExpressions: readonly GeneratedExpression[]
```
charlierudolph, line 14: Why was readonly added here?

aslakhellesoy: I assume to prevent accidental modification of the array and to signal that the array is not meant to be modified. I tend to use it in most signatures and only use modifiable arrays in implementations.

features/step_definitions/message_steps.ts
davidjgoss: (it does just compile down to the ordinals sadly)

features/attachments.feature
davidjgoss: (this was done)

src/formatter/step_definition_snippet_builder/snippet_syntax.ts
davidjgoss: Yep - the `cucumber-expressions` library declares it as a `ReadonlyArray` as of https://github.com/cucumber/cucumber/commit/1b1cde5b59d79e155b8324701c46685a45b0dd2f, so we have to match that here to not get a compiler error.

src/models/data_table.ts
davidjgoss: To support the `transpose` method, added via https://github.com/cucumber/cucumber-js/pull/1318/commits/7f746ae524e452cf25c60c5380354b7166520e19 - this is present in other data table impelmentations I think

davidjgoss:
@charlierudolph @aslakhellesoy sorry this has taken awhile to come together - it's been a weird few months.

To recap on where we are now in terms of CCK:

- We're now using the latest versions of all the monorepo libraries (as of today)
- Looking at the actual output vs the CCK fixtures, it basically matches except for ids, times and some ordering (I think we could get it to line up perfectly if we _really_ wanted to by changing the order of some things here, but that's non-trivial and wouldn't seem to warrant holding up a release for
- Last step is to test with the HTML formatter and/or push the actual outputs to Cucumber Reports

And in terms of getting this PR green, a few of our existing acceptance tests are failing:

- URIs are coming out as absolute rather than relative to the cwd, I'm looking at that now
- Some durations have changed, looking at that next
- `gherkin_parse_failure.feature` is failing - it looks like Gherkin doesn't throw any more for a malformed feature - @aslakhellesoy do I have that right and if so is that intentional?

charlierudolph:
Awesome have the build is now passing! Is the `compatibility` folder meant to stay checked in or was that added temporarily and should go away before merging this PR?

davidjgoss:
@charlierudolph yeah supposed to stay in for now, though it's not great. Ideally if the CCK is [published at `@cucumber/compatibility-kit`](https://www.npmjs.com/package/@cucumber/compatibility-kit) then we can just use the features and fixtures from there. We could also checkout the tag from the monorepo but seems less neat. It's not running as a test in CI yet as there are still quite a lot of differences, but it's really all ids and ordering.

davidjgoss:
@charlierudolph also about to look at adding `html-formatter` and/or `json-formatter` as `devDependencies` and piping the cck output to them as a sort of e2e test.

aslakhellesoy:
Hey @davidjgoss - I've released `@cucumber/compatibility-kit` version `2.0.0`. It's got all the latest and greatest.

aslakhellesoy:
For Cucumber-Ruby and Cucumber-JVM we implemented some logic to compare the generated messages with those from the compatibility kit, ignoring differences in IDs, timestamps, etc.

* https://github.com/cucumber/cucumber-jvm/blob/main/compatibility/src/test/java/io/cucumber/compatibility/CompatibilityTest.java
* https://github.com/cucumber/cucumber/tree/master/compatibility-kit/ruby/lib
  * Ideally this should be moved from the monorepo to cucumber-ruby

davidjgoss:
Thanks @aslakhellesoy! I'll start hooking that up now.

And yes, I'll follow the pattern of normalising ids, timestamps etc - I think there are some existing utilities in cucumber-js for that I can lean on. The only other mismatch would be ordering - in the cck fixtures, all the `testCase` messages are emitted together at the start of the overall test run, where in cucumber-js they're emitted before each test case runs. I might be able to make the assertion ignore ordering, and perhaps that's a refactor we can circle back to.

aslakhellesoy:
That sounds like a good plan.

The message protocol doesn't currently define message ordering. It's currently only informally constrained by consumers' ability to consume messages with different ordering. The most important message consumers are:

* https://github.com/cucumber/cucumber/tree/master/html-formatter
* https://github.com/cucumber/cucumber/tree/master/json-formatter

So in the absence of a formal verification of message ordering, send the messages generated by your cck implementation to those formatters as part of the build. If it doesn't crash, and a cursory eye-balling suggests the result is OK, then we can be fairly confident that "it works".

We did experiment with a more formal verification of message ordering in #642 and #656 but that never landed...

src/cli/helpers.ts
```diff
@@ -195,7 +195,8 @@ function emitTestCaseHooks(
   ;[]
     .concat(
       supportCodeLibrary.beforeTestCaseHookDefinitions,
-      supportCodeLibrary.afterTestCaseHookDefinitions
+      // order has been reversed before - need it back in original order
```
charlierudolph, line 198: Can we solve this in a different way by having the support code library keep it in the original order and only reverse it in the runtime when we are about to execute them?

davidjgoss: done

davidjgoss:
Okay, we now have a script as part of the `test` script that runs the `html-formatter` against most of the CCK output (there are some issues with duplicate step names etc so haven't done all right now). There's an issue with the formatter when installed as a local dependency rather than globally, which is coming from [this line](https://github.com/cucumber/cucumber/blob/master/html-formatter/javascript/src/cli-main.ts#L33) - the CSS file doesn't resolve. I'll open a separate issue for that and look at fixing it, but in the meantime I've worked around using `npx` which does a temporary global install. The formatted output looks good!

The only outstanding issue now is with undefined parameter types. Previous `cucumber-js` behaviour was to throw immediately when trying to define a step with an undefined parameter type, meaning the actual test run never happens. Whereas per the CCK, the test run should still happen and a message for the undefined parameter type should be emitted. It would be good to clarify, but my current thinking is:

- Do the test run, but exit with code 1 if there were undefined parameter types
  - except if using `fail-fast` option, in which case exit immediately after emitting the support code messages?
- Add output to the summary and progress-bar formatters (fixing the current failing feature test)

Does that sound okay @charlierudolph @aslakhellesoy?

aslakhellesoy:
That all sounds good to me David.

davidjgoss:
And we are green.

davidjgoss:
@charlierudolph good to merge?

charlierudolph:
Awesome having cck and formatter tests passing! Let me take a look.

compatibility/cck_spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  const testCases: any[] = messages.filter((message) =>
+    doesHaveValue(message.testCase)
+  )
+  const everythingElse: any[] = messages.filter((message) =>
+    doesNotHaveValue(message.testCase)
+  )
+  const testRunStarted = everythingElse.findIndex((message) =>
+    doesHaveValue(message.testRunStarted)
+  )
+  // move all `testCase` messages to just after `testRunStarted`
```
charlierudolph, line 94: Should we create an issue to switch to doing this as expected by the cck?

compatibility/cck_spec.ts
```diff
@@ -0,0 +1,97 @@
...
+          'pickleStepId',
+          'stepDefinitionIds',
+          'testCaseId',
+          'testCaseStartedId',
+          'testStepId',
+          // time
+          'nanos',
+          'seconds',
+          // errors
+          'message',
```
charlierudolph, line 73: Are we expecting all of these to be excluded long term? I would hope with predictable ids that big block could be fixed.  Do you think its worth having an issue to try to drive this list down.

compatibility/features/attachments/attachments.ts
```diff
@@ -0,0 +1,75 @@
+import { Before, When, World } from '../../..'
+import { ReadableStreamBuffer } from 'stream-buffers'
+import fs from 'fs'
+import path from 'path'
+
+Before((): void => undefined)
```
charlierudolph, line 6: Why is this here?

cucumber.js
```diff
@@ -1,13 +1,43 @@
...
+  'message',
+].join(' ')
+
+const FORMATTERS_INCLUDE = [
+  'attachments',
+  'data-tables',
+  'minimal',
+  'parameter-types',
+  'rules',
+]
```
charlierudolph, line 25: How was this list formed?

package.json
```diff
@@ -222,6 +229,7 @@
     "eslint-plugin-promise": "4.2.1",
     "eslint-plugin-standard": "4.0.1",
     "fs-extra": "9.0.1",
+    "globby": "^11.0.0",
```
charlierudolph, line 232: `glob` is already a dependency. Can we use that?

src/runtime/pickle_runner.ts
```diff
@@ -243,6 +277,11 @@ export default class PickleRunner {
       this.result.status === Status.FAILED &&
       attempt + 1 < this.maxAttempts
     ) {
+      /*
+      TODO dont rely on `testStepResult.willBeRetried`, it will be moved or removed
+      see https://github.com/cucumber/cucumber/issues/902
+       */
```
charlierudolph, line 283: Can we make a cucumber-js issue for this?

src/runtime/stopwatch.ts
charlierudolph: Not seeing any tests for this.

charlierudolph:
Time reporting looks a bit off:

```
⋊> ~/p/cucumber-js on cck ◦ time node ./bin/cucumber-js
...
239 scenarios (239 passed)
1201 steps (1201 passed)
0m00.449s
       38.04 real        23.48 user         3.40 sys
```

src/runtime/stopwatch.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return current
+  }
+
+  timestamp(): messages.ITimestamp {
+    const current = this.duration()
+    return {
+      nanos: current.nanos(),
```
charlierudolph, line 42: `nanos()` returns the whole duration in nanos whereas I think we want only the second fraction. 

Along the same line `seconds()` returns a fraction whereas I think we want a whole number.

https://github.com/joeledwards/node-durations#methods-1

compatibility/cck_spec.ts
davidjgoss: Sure, raised #1408

compatibility/cck_spec.ts
davidjgoss: I agree - when I was working against a slightly earlier CDK version I think it was only the `testCase` ordering that made the ids not line up.

But cck has [changed](https://github.com/cucumber/cucumber/commit/e254f6a9ba2c41e52485a76c1f78221cef63879a#diff-102d4a7789d087b284dbfed6836597fd) from using predictable ids to uuids so not sure where that leaves us.

compatibility/features/attachments/attachments.ts
davidjgoss: From the reference steps in cck:

```typescript
// Cucumber-JVM needs to use a Before hook in order to create attachments
Before(() => undefined)
```

Have to include it here too otherwise we're a bunch of messages out.

cucumber.js
davidjgoss: The ones that don't have conflicting step definitions, and don't cause everything else to fail. So everything except `hook` and `undefined-parameter-types` (realised I had missed two others, pushed a fix).

src/runtime/pickle_runner.ts
davidjgoss: Yep raised #1409

package.json
davidjgoss: Good point, done.

src/runtime/stopwatch.ts
davidjgoss: Right, duh, I’ll fix this and have a look at the timing mismatch too.

compatibility/cck_spec.ts
aslakhellesoy: The predictable ids and timestamps are a dead end. The cck no longer uses them.

It's a dead end because various Cucumber implementations emit messages in slightly different order than what `fake-cucumber` (the engine behind the cck) emits. It would be too much work to make them all emit in the same order.

So what we opted for instead is a validation mechanism where certain values (ids, timestamps) are not compared. We pair this with smoke tests that send the generated messages through 2 of the most advanced message consumers we have - cucumber-html and json-formatter. If these don't crash and a cursory eye-balling looks OK, we're confident enough that the emitted messages are ok.

compatibility/features/attachments/attachments.ts
aslakhellesoy: Yep, that's correct.

src/runtime/stopwatch.ts
davidjgoss: Okay, pushed a fix for the timestamp seconds/nanos.

With the durations, when testing locally I can see the same behaviour on this branch and master - the summary formatter's total duration is on average about 67-68% of the actual elapsed time. Agree with your theory, it's the accumulation of user code without the cucumber plumbing in between. Whereas `testRunFinished.timestamp` is more because it's the actual time elapsed from when we emit `testRunStarted` (confirmed this is accurate with some logging to make sure).

On this basis should we change the summary formatter to use the same result as the message?

davidjgoss: Thanks for sorting the timestamps! Think we are finally there now

davidjgoss:
Will hit merge on this in the morning if no other issues

Issue #1423, opened 2020-09-01
Add --publish option


aslakhellesoy:
Cucumber-JVM and Cucumber-Ruby now have a --publish option. More details here:

- https://github.com/cucumber/cucumber-jvm/blob/main/release-notes/v6.5.0.md
- https://github.com/cucumber/cucumber-ruby/blob/master/release-notes/5.0.0.md

We should implement this for Cucumber-js too. I'm happy to start on that with @sebrose and anyone else who wants to help out.

It's always hard to estimate how long things take, but we might be able to get it done this week. What are the current plans for a release? Is master ready to go? /cc @davidjgoss @charlierudolph

davidjgoss:
@aslakhellesoy I'm happy to help out with that as and when.

Regarding a 7.0.0, I think we just need to finish #1414, then merge #1411 and good to go - there may be something I'm missing though @charlierudolph?

aslakhellesoy:
Excellent! I'd love to include #1423 / #1424 in the 7.0.0 release. I think I'll be spending most of the time this week working on that. I'll update the PR with TODOs - then we can discuss how to split up the work.

PR #1424, opened 2020-09-01
Publish option

aslakhellesoy:
This implements #1423 

- [x] Print a banner when `--publish` is not specified
- [x] Add a mechanism that allows users to silence the banner with a file
- [x] Implement the logic for sending message formatter output to a HTTP URL
  - See https://github.com/cucumber/cucumber-ruby/pull/1464 and https://github.com/cucumber/cucumber-jvm/pull/2099
- [x] Allow env var overrides as specified in https://github.com/cucumber/cucumber-jvm/blob/main/release-notes/v6.5.0.md
- [x] Print a banner of where the report can be accessed
- [x] Create Cucumber-JVM / Cucumber-Ruby issues to use the server-provided banner as we do here


aslakhellesoy:
We have liftoff! Check this out: https://reports.cucumber.io/reports/7d8ac0d5-738b-4250-bb02-02feaea95aa7

davidjgoss:
Nice!

Just noticed it reports as having been run 51 years ago - are we doing some thing wrong with time stamps?

aslakhellesoy:
Yes I guess so! I’ll have a look at that tomorrow.

dependency-lint.yml
```diff
@@ -19,7 +19,8 @@ executedModules:
 ignoreErrors:
   missing: []
   shouldBeDependency: []
-  shouldBeDevDependency: []
+  shouldBeDevDependency:
+    - tmp # used in http_stream.ts as well as in test code
```
charlierudolph, line 23: If this is now used in source code we should move this from a dev dependency to a dependency and remove the need to ignore the errory

src/cli/publish_banner.ts
```diff
@@ -0,0 +1,18 @@
...
+${bold}${green}│${reset} Command line option:    ${cyan}--publish${reset}                                        ${bold}${green}│${reset}
+${bold}${green}│${reset} Environment variable:   ${cyan}CUCUMBER_PUBLISH_ENABLED${reset}=${cyan}true${reset}                    ${bold}${green}│${reset}
+${bold}${green}│${reset}                                                                          ${bold}${green}│${reset}
+${bold}${green}│${reset} More information at ${underline}${bold}${cyan}https://reports.cucumber.io/docs/cucumber-js${reset}         ${bold}${green}│${reset}
+${bold}${green}│${reset}                                                                          ${bold}${green}│${reset}
+${bold}${green}│${reset} To disable this message, add this to your ${bold}./cucumber.js${reset}:                 ${bold}${green}│${reset}
+${bold}${green}│${reset} ${bold}module.exports = { default: '--publish-quiet' }${reset}                          ${bold}${green}│${reset}
+${bold}${green}└──────────────────────────────────────────────────────────────────────────┘${reset}
```
charlierudolph, line 17: This looks tough to maintain. Let me take a look at updating this to use libraries for the ANSI fonts and the table border.

features/publish.feature
```diff
@@ -0,0 +1,131 @@
...
+      ┌──────────────────────────────────────────────────────────────────────────┐
+      │ Share your Cucumber Report with your team at https://reports.cucumber.io │
+      │                                                                          │
+      │ Command line option:    --publish                                        │
+      │ Environment variable:   CUCUMBER_PUBLISH_ENABLED=true                    │
+      │                                                                          │
+      │ More information at https://reports.cucumber.io/docs/cucumber-js         │
+      │                                                                          │
+      │ To disable this message, add this to your ./cucumber.js:                 │
+      │ module.exports = { default: '--publish-quiet' }                          │
```
charlierudolph, line 103: I feel conflicted about this. This feels annoying to me as a user. Certainly glad we implemented the CLI option to suppress it but why does this need to be outputted a user runs the tool without publish?

Could this be a message shown on install / the help page / somewhere visible on the README?

src/cli/publish_banner.ts
charlierudolph: Updated. Note removed the bold as without it, it looks a bit better on my machine (top bold, bottom non-bold). Let me know if we really want to keep the bold:

Screen Shot 2020-09-07 at 3 55 52 PM

dependency-lint.yml
charlierudolph: Removed as appears unnecessary since tmp is already moved to a dependency

test/fake_report_server.ts
```diff
@@ -0,0 +1,97 @@
...
+  public receivedHeaders: http.IncomingHttpHeaders = {}
+
+  constructor(private readonly port: number) {
+    const app = express()
+
+    app.put('/s3', (req, res) => {
+      this.receivedHeaders = { ...this.receivedHeaders, ...req.headers }
+
+      // eslint-disable-next-line @typescript-eslint/no-this-alias
+      const server = this
```
charlierudolph, line 26: Why is `server` needed? Couldn't `write` be defined with an arrow function?

test/fake_report_server.ts
```diff
@@ -0,0 +1,97 @@
...
+      const captureBodyStream = new Writable({
+        write(chunk: Buffer, encoding: string, callback: Callback) {
+          server.receivedBodies = Buffer.concat([server.receivedBodies, chunk])
+          callback()
+        },
+      })
+
+      pipeline(req, captureBodyStream, (err) => {
+        if (err !== null && err !== undefined)
```
charlierudolph, line 35: We have a helper for this `doesHaveValue` in src/value_checker

test/fake_report_server.ts
```diff
@@ -0,0 +1,97 @@
...
+          new Promise((resolve, reject) => {
+            if (socket.destroyed) return resolve()
+            socket.on('close', resolve)
+            socket.on('error', reject)
+          })
+      )
+    )
+    return new Promise((resolve, reject) => {
+      this.server.close((err) => {
+        if (err !== null && err !== undefined) return reject(err)
```
charlierudolph, line 88: Can use `doesHaveValue` here as well

src/formatter/http_stream.ts
```diff
@@ -0,0 +1,139 @@
...
+    })
+  }
+
+  private sendRequest(
+    url: string,
+    method: HttpMethod,
+    callback: (err: Error | null | undefined, url?: string) => void
+  ): void {
+    // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+    const httpx = url.match(/^https:/) ? https : http
```
charlierudolph, line 79: Could remove the need for disabling the strict boolean expressions by wrapping `url.match` with `doesHaveValue`

src/formatter/http_stream.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // TODO: Follow regular 3xx redirects
```
charlierudolph, line 81: Is this expected to be done in this PR? Also may be better to use a npm library for this that follows redirects automatically.

https://github.com/visionmedia/superagent is one example that will follow redirects

sebrose: We do not currently support 3xx redirects and are not intending to implement this functionality in this PR.

We'll remove the TODO

dependency-lint.yml
aslakhellesoy: Resolving this since @charlierudolph fixed it in 841d8d6b4ffef2f5a4917fd97e655f4d4a575258

src/cli/publish_banner.ts
aslakhellesoy: Nice thanks!

features/publish.feature
aslakhellesoy: We've already implemented it this way in Cucumber-Ruby and Cucumber-JVM. So far we've not heard any complaints, just positive feedback.

We want as many people as possible to discover this new functionality, and showing it after the run will reach more people than if it's only shown on install and in documentation (which many people never look at).

Let's see if anyone complains.

test/fake_report_server.ts
aslakhellesoy: Fixed in 508c876b05bee840466de3845becf87e27f79001

test/fake_report_server.ts
aslakhellesoy: Fixed in 9629395a0ab37f0bf59518e6c68e970d264c3360

test/fake_report_server.ts
aslakhellesoy: Fixed in deprecated and will be removed in the next majo

aslakhellesoy: Fixed in 9629395a0ab37f0bf59518e6c68e970d264c3360

src/formatter/http_stream.ts
aslakhellesoy: Fixed in 9629395a0ab37f0bf59518e6c68e970d264c3360

src/cli/argv_parser.ts
```diff
@@ -168,6 +170,16 @@ const ArgvParser = {
         'Use predictable ids in messages (option ignored if using parallel)',
         false
       )
+      .option(
+        '--publish',
+        'Publish a report to https://reports.cucumber.io',
+        false
+      )
+      .option(
+        '--publish-quiet',
```
charlierudolph, line 179: Note, I think this is named wrong. This initially makes me think it suppresses output when using the publish option instead of suppressing an advertisement about publish. Given already used in a ruby/jvm this is fine for now but in long run would prefer to rename this to something like `--publish-suppress-advertisement`

src/cli/configuration_builder.ts
```diff
@@ -157,15 +163,43 @@ export default class ConfigurationBuilder {
...
   getFormats(): IConfigurationFormat[] {
     const mapping: { [key: string]: string } = { '': 'progress' }
     this.options.format.forEach((format) => {
       const [type, outputTo] = OptionSplitter.split(format)
       mapping[outputTo] = type
     })
+    if (this.isPublishing()) {
+      const publishUrl =
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+        process.env.CUCUMBER_PUBLISH_URL || DEFAULT_CUCUMBER_PUBLISH_URL
```
charlierudolph, line 190: For this, we also have a helper is src/value_checker.

`valueOrDefault(a, b)` which returns a if not null / undefined, otherwise b

Using that should be able to remove the eslint disable

davidjgoss: Not sure but this newer syntax might work too:

```typescript
process.env.CUCUMBER_PUBLISH_URL ?? DEFAULT_CUCUMBER_PUBLISH_URL
```

src/cli/argv_parser.ts
vincent-psarga: Yes we can change it in the future but then we'll need to do so in all implementations.

src/cli/configuration_builder.ts
vincent-psarga: resolved by 860f430b6bc5235e51ef4dc51318556f68ef550c

features/publish.feature
michaelm-rsi: Me. I'll complain.  I'm beginning the setup of a test suite and this message is annoying to look at.  I tried dropping the module.exports into my step definition file but it didn't get rid of this message. So how do I get rid of this?

aslakhellesoy: > I tried dropping the module.exports into my step definition file

You're not supposed to drop that in your step definition files, but in your global `cucumber.js` [config file](https://github.com/cucumber/cucumber-js/blob/master/docs/cli.md#profiles) at the root of your project.

How can we make this harder to misunderstand @michaelm-rsi?

> I'm working on a corporate project which will NEVER publish test results onto a public server for security reasons.

I completely understand that, which is why we've added an option to turn this off. We're also considering making Cucumber Reports available as a server you can install on-prem.

Issue #1309, opened 2020-04-21
--name does not recognize a scenario if special characters are used in the scenario name


abhinaba-ghosh:
Hi,
My scenario name is something like:

```
Scenario: (Hello) This is my first test
```
When I am trying to run the scenario with CLI  command as:
```
cucumber-js test/features --name='(Hello) This is my first test'
```
This is not able to recognize and run the scenario. I am using cucumber 6.x

If I just delete the braces, then it is able to recognize and run the scenario. 
Any guidance will help!

kamolins:
@davidjgoss 

Not sure if this is a valid solution, but changing `match` to `includes` [here](https://github.com/cucumber/cucumber-js/blob/3e0c21174fc86603f452106543912e52935e9341/src/pickle_filter.ts#L120) should fix the issue. 
If this is an acceptable fix, I can make a PR!

davidjgoss:
Hi @abhinaba-ghosh, the `--name` argument treats the input as a regular expression, so if you have characters like parentheses you should be able to escape them. For example:

```shell
--name='\(Hello\) This is my first test'
```

Does that work?

abhinaba-ghosh:
It works! Thanks for the help.

PR #1445, opened 2020-09-22
Unit test for CLI flag --name + doc update

kamolins:
~~Addresses: #1309~~ 

- Added unit test to ensure that CLi flag --name supports regex
- Updated docs


davidjgoss:
Thanks for doing this @kamolins, looks good.

Would you be able to add a test that fails without your change and passes with it?

kamolins:
@davidjgoss Done!

davidjgoss:
@kamolins sorry, looking again at [the documentation](https://github.com/cucumber/cucumber-js/blob/master/docs/cli.md#running-specific-features), this argument is listed as supporting regular expressions, so moving away from `match` will break that which I don't think we want to do.

You should be able to escape the parentheses and other characters that have regex meaning with backslashes, I think?

I do still think there is worthwhile change to make here if you wanted to pivot this PR a bit:
- The doc briefly mention regex, but it might be worth showing an example with escaping
- No tests broke when you changed from `match` to `include` - we should add one that proves the regex support

kamolins:
@davidjgoss You make a good point! `includes` won't match a regex, afaik. I should have read the docs more carefully. I will update the PR.

kamolins:
@davidjgoss Updated the PR. Lmk if I should do anything else.

By the way, my previous push was marked as failed [here](https://ci.appveyor.com/project/charlierudolph/cucumber-js/builds/35585358/job/srgvx68bv6249r2e) on node v10 with an unrelated issue. Not really sure what went wrong there.

davidjgoss (approved):

aslakhellesoy:
Hi @kamolins,

Aslak Hellesøy
Creator of Cucumber

PR #1446, opened 2020-09-24
Proposal to address issue #1079

kamolins:
Howdy!

I wanted to tackle #1079 but I wasn't sure where to start or how to go about it so I made a proposal PR to add function names to hooks.

image

Keep in mind that this is a prototype hence code nor implementation is meant to be final. (only updated relevant parts for this proposal)

I would love to hear feedback on this, and in case there is a better/different way to implement this as to achieve the same goal of identifying hooks in some way I am willing to tackle this if you can provide me some pointers since I am new to this.

### DIFF
--- a/features/ambiguous_step.feature
+++ b/features/ambiguous_step.feature
@@ -1,5 +1,12 @@
 Feature: Ambiguous Steps
 
+  # Left here for testing purposes
+  @test
+  Scenario: Test this
+    Given Cake exist
+    When I eat cake
+    Then My belly is full
+
   Scenario:
     Given a file named "features/a.feature" with:
       """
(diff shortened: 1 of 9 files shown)


kamolins:
Going to close this for the time being and wait till 7.0.0 is released

davidjgoss:
@kamolins sorry for the late reply. 7.0.0 is out now so it might be good to revisit this.

One thought from me: I think it would be nice to be able to name the hook in plain language e.g. "set up test data" or "initialise browser session" or "collect logs" etc. We could potentially allow a name with a non-breaking change to the `IDefineTestCaseHookOptions` signature, so something like:

```js
Before({name: "initialise browser session"}, async function() {
  // do stuff
});
```

What do you think?

kamolins:
I think your approach is great. I am definitely not familiar enough with the codebase to see if there are any implications or downsides to this, but in general, I like it a lot more than what I proposed.
When I have some time, I will rebase this PR and see if I can actually implement it.

PR #1476, opened 2020-11-28
docs: add clarity to docs on `--require`

### LINKED ISSUE #1460: Insufficient documentation in CLI.md
In the docs currently:

> Use --require <GLOB|DIR|FILE> to require support files before executing the features. Uses glob patterns. If not used, the following files are required:
> 
> If the features live in a features directory (at any level)
> features/**/*.js
> Otherwise
> <DIR>/**/*.js for each directory containing the selected features
> Automatic loading is disabled when this option is specified, and all loading becomes explicit.

So how the HELL are we supposed to do explicit loading??  That's documented nowhere.

davidjgoss:
- note that it supports multiple usage
- make clear that defaults are not applied once you specify with the option

Closes #1460

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -28,14 +28,17 @@ needs to be required in your support files and globally installed modules cannot
 
 ## Requiring support files
 
-Use `--require <GLOB|DIR|FILE>` to require support files before executing the features. Uses [glob](https://github.com/isaacs/node-glob) patterns.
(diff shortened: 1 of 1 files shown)


davidjgoss:
One for another PR, but wonder if we should also pull the section on `--require-module` up to be right after this section?

charlierudolph (approved):

PR #1505, opened 2020-12-05
Revert "Configure WhiteSource for GitHub.com"

mpkorstanje:
Reverts cucumber/cucumber-js#1490

@mloewinger74 please ensure the build passes before merging any changes. This should include upgrading vulnerable dependencies. Additionally it would be appreciated if you could discuss any sweeping changes in the #commiters slack first.


aslakhellesoy:
Yes absolutely. I should have verified that this didn't break the build. My bad. I'll try again and make sure all is green before I try again.

PR #1510, opened 2020-12-11
build: bump cucumber library dependencies to latest

davidjgoss:


yarn.lock
```diff
@@ -212,7 +212,7 @@
   dependencies:
     "@cucumber/messages" "^13.0.1"
 
-"@cucumber/gherkin@^15.0.0", "@cucumber/gherkin@^15.0.2":
+"@cucumber/gherkin@^15.0.0":
```
aslakhellesoy, line 215: It's a bit suspicious having both `^15` and `^16` here. I think we should only have `^16`. Let's find out what that's about.

aslakhellesoy (changes requested):

yarn.lock
aslakhellesoy: We need to make a new release of `gherkin-utils`. The latest release (`2.1.0`) still depends on `@cucumber/gherkin` version  `15`. I'll make a new release of that lib (it won't have a dependency on `@cucumber/gherkin` at all)

davidjgoss: Okay thanks

aslakhellesoy:
We need to make a new release of `@cucumber/react` and `@cucumber/html-report` after https://github.com/cucumber/cucumber/issues/1275 is fixed. And then update this PR with the new versions.

davidjgoss:
@aslakhellesoy okay thanks for update.

If we're looking at react and html-formatter anyway would be good to squeeze https://github.com/cucumber/cucumber/pull/1260 in (I'll add some screenshots to the PR, thought I had done that already).

(Happy to pick up markdown library change if needed.)

aslakhellesoy:
Yes indeed @davidjgoss - we'll merge as many react/html related issues as we can :-)

davidjgoss:
@aslakhellesoy we do still have a couple of things in the lockfile:

- `@cucumber/gherkin` 15.0.0 still being brought in via `@cucumber/query` 7.0.0
- They also both bring in `@cucumber/messages` 13.0.1 (we have moved onto 13.2.1)

So it looks like to resolve that fully we'd need to bump the deps in `query`, then release `react` and `html-formatter` with updated dep on that.

Wondering if we should look at making some of these things `peerDependencies` with a sensible range?

davidjgoss:
(Looking at history for `query` I think the dependency on `gherkin` is already gone and it just needs a release.)

davidjgoss:
Also seeing an error with html-formatter results, need to look at that.

PR #1549, opened 2021-01-23
Add missing @cucumber/react

mucsi96:
While using the html formatter I am getting an error of missing `cucumber-react.css` CSS file from `@cucumber/react` package. I see based on code that it's expected to be installed. But it's not added to the list of dependencies.

https://github.com/cucumber/cucumber-js/blob/master/src/formatter/html_formatter.ts#L15

### DIFF
--- a/package.json
+++ b/package.json
@@ -173,6 +173,7 @@
     "@cucumber/html-formatter": "^11.0.4",
     "@cucumber/messages": "^13.2.1",
     "@cucumber/query": "^7.0.1",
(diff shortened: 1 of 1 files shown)


davidjgoss:
@mucsi96 cucumber-js depends on `@cucumber/html-formatter` which in turn depends on `@cucumber/react` - if you don't find it in `node_modules` it suggest an error with your install. If the package _is_ installed and you still see this error, could you raise an issue with a link to a minimal repro?

PR #1575, opened 2021-02-09
Remove `any` in `IJsonStep` interface `arguments` property

wafuwafu13:
Hi :)
I resolved `TODO`
https://github.com/cucumber/cucumber-js/blob/625fab034eea768bf74f7a46993a57182204ddf6/src/formatter/json_formatter.ts#L53

### DIFF
--- a/src/formatter/json_formatter.ts
+++ b/src/formatter/json_formatter.ts
@@ -50,7 +50,7 @@ export interface IJsonScenario {
 }
 
 export interface IJsonStep {
-  arguments?: any // TODO
+  arguments?: {} | []
   embeddings?: any // TODO
   hidden?: boolean
   keyword?: string // TODO, not optional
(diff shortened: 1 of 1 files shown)


src/formatter/json_formatter.ts
```diff
@@ -50,7 +50,7 @@ export interface IJsonScenario {
 }
 
 export interface IJsonStep {
-  arguments?: any // TODO
+  arguments?: {} | []
```
aurelien-reeves, line 53: It looks like arguments is always an array
Why having the `{} |`?

src/formatter/json_formatter.ts
```diff
@@ -129,12 +138,12 @@ export default class JsonFormatter extends Formatter {
   formatStepArgument(
     stepArgument: messages.IPickleStepArgument,
     gherkinStep: messages.GherkinDocument.Feature.IStep
-  ): any {
+  ): [] | {} {
```
aurelien-reeves, line 141: This method actually always return an array
Why `| {}`?

aurelien-reeves (commented):
Thanks for your contribution! :)

Could someone more experienced with cucumber-js take a look too?
Apart for my comments, it looks pretty good to me

Cc. @davidjgoss

davidjgoss (changes requested):
Thanks again for contributing.

To pick up on @aurelien-reeves comments, typing as `{} | []` isn't really any narrower than `any`.

Digging through the code a little, it looks like the value will be an array with each item being a formatter-specific representation of either a doc string or a data table. See the `formatDocString` and `formatDataTable` methods in the same file - I think if you start by getting a specific type on both of those you should be able carry that up from there.

aurelien-reeves:
Thank you for your contribution
We want stricter typing than what you have proposed

See. https://github.com/cucumber/cucumber/issues/1138 which would apply to cucumber-js too

PR #1587, opened 2021-02-26
emit relative urls

davidjgoss:


aurelien-reeves:
@davidjgoss would you have any info about this one?

davidjgoss:
@aurelien-reeves  ah thanks for the nudge, this can close actually, needs a different approach

PR #1607, opened 2021-03-05
build: inline sources in source maps

davidjgoss:
We were already emitting source maps, but Node was unable to really use them at runtime because we don't also include the TypeScript source files in the package. Doing that feels a bit off to me - could lead to issues with end users trying to use the source files directly with their own tsconfig. This change means we inline the source content with the source map files, so stack traces and debuggers can do their thing.


charlierudolph (approved):
Sounds good. Thanks for the explanation!

PR #1608, opened 2021-03-09
Display error banner when Reports server returns status code ≠ 2xx

vincent-psarga:
Other Cucumber implementations display the content send by the server upon failure. For example, with Cucumber ruby:

```
$ CUCUMBER_PUBLISH_TOKEN=plop bundle exec cucumber features/docs/getting_started.feature
WARNING: Nokogiri was built against LibXML version 2.9.10, but has dynamically loaded 2.9.4
Using the default profile...
........

2 scenarios (2 passed)
8 steps (8 passed)
0m0.023s
┌──────────────────────────────────────────────────────────────┐
│ Invalid CUCUMBER_PUBLISH_TOKEN: plop                         │
│ Go to https://reports.cucumber.io/profile to get your token. │
└──────────────────────────────────────────────────────────────┘
Traceback (most recent call last):
	1: from /Users/vincent.pretre/dev/open-source/cucumber-ruby-meta/cucumber-ruby/lib/cucumber/formatter/io.rb:40:in `block (3 levels) in new'
/Users/vincent.pretre/dev/open-source/cucumber-ruby-meta/cucumber-ruby/lib/cucumber/formatter/http_io.rb:79:in `close': request to https://messages.cucumber.io/api/reports failed with status 401 (StandardError)
```

This is an attempt to get the same behavior in `cucumber-js`

TODO:
 - [x] Add a example of what's excepted
 - [x] Display the server response
 - [x] Fail when the server does not return 2xx
 - [ ] Solve the `null` line displayed above the banner

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,6 +14,10 @@ jobs:
           - ubuntu-latest
           - windows-latest
         node-version: [10.x, 12.x, 14.x, 15.x]
+      # This could be disabled before merging PR, but honestly, it's a bit frustrating
+      # not having an overview of what works and what don't as soon as one job fails.
+      fail-fast: false
+
     steps:
     - uses: actions/checkout@v2
     - name: with Node.js ${{ matrix.node-version }} on ${{ matrix.os }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,8 +14,6 @@ jobs:
           - ubuntu-latest
           - windows-latest
         node-version: [10.x, 12.x, 14.x, 15.x]
-      # This could be disabled before merging PR, but honestly, it's a bit frustrating
-      # not having an overview of what works and what don't as soon as one job fails.
       fail-fast: false
 
     steps:
--- a/features/publish.feature
+++ b/features/publish.feature
@@ -112,7 +112,7 @@ Feature: Publish reports
       │ Error invalid token │
       └─────────────────────┘
 
-      Error: Unexpected http status 401 from GET http://localhost:9987
+      Unexpected http status 401 from GET http://localhost:9987
       """
 
   @spawn
--- a/features/publish.feature
+++ b/features/publish.feature
@@ -54,7 +54,7 @@ Feature: Publish reports
 
   @spawn
   Scenario: Report is published when CUCUMBER_PUBLISH_TOKEN is set
-    When I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_TOKEN=keyboardcat`
+    When I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_TOKEN=f318d9ec-5a3d-4727-adec-bd7b69e2edd3`
     Then it passes
     And the server should receive the following message types:
       | meta             |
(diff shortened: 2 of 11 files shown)


vincent-psarga:
The build is not really stable (I managed to get it green by relaunching - and after it failed again).

Apparently that comes from some tests I have added and the way the stream errors are handled. I'm going to dig a bit more into that (although the solution I have in mind is not that beautiful to be honest :D )

aslakhellesoy:
@vincent-psarga are the non-deterministic tests new tests in this PR? I can help have a look at this.

davidjgoss:
The flaky tests I think started from the original change where publish support was added per #1457

I tried a few things a while ago but hit a bit of a wall tbh.

.github/workflows/build.yml
```diff
@@ -14,6 +14,10 @@ jobs:
           - ubuntu-latest
           - windows-latest
         node-version: [10.x, 12.x, 14.x, 15.x]
+      # This could be disabled before merging PR, but honestly, it's a bit frustrating
+      # not having an overview of what works and what don't as soon as one job fails.
+      fail-fast: false
```
davidjgoss, line 19: Agree with keeping this!

vincent-psarga: great :) I'll remove the comment then :)

vincent-psarga:
> @vincent-psarga are the non-deterministic tests new tests in this PR? I can help have a look at this.

At least from what I see, it's the test that I've introduced which is flaky. I'm looking at a way to not spawn Cucumber in order to fix this (worked fine with Ruby getting rid of spawning, so I give a try)

aurelien-reeves:
Just tried to remove the delays in the server
The error went back

So it seems the combination of the delays, and the ports, stabilized the tests.

aurelien-reeves:
@davidjgoss do you have any new feedback regarding that PR?
We would like to merge it and release it.

I plan to open a new PR dedicated to the inconsistent test once that one is merged.

davidjgoss:
@aurelien-reeves I'll review this morning, thanks for the nudge!

features/publish.feature
```diff
@@ -101,6 +101,20 @@ Feature: Publish reports
       │ module.exports = { default: '--publish-quiet' }                          │
       └──────────────────────────────────────────────────────────────────────────┘
       """
+
+  @spawn
+  Scenario: when results are not published due to an error raised by the server, the banner is displayed
+    When I run cucumber-js with env `CUCUMBER_PUBLISH_TOKEN=keyboardcat`
+    Then it fails
```
davidjgoss, line 108: As a general point, not sure how users would feel about their CI pipeline failing because e.g. cucumber reports is down or they got their token wrong (but all tests passed). We could wait for feedback and look at making it configurable via formatter options if needed.

davidjgoss (approved):
LGTM

features/publish.feature
vincent-psarga: Well, that's already the case if you publish with a wrong token, `cucumber-js` exits with code 1. What this PR does is just providing extra information.
And I think it should fail if the publication did not succeed, as something went wrong. It a bit like generating a JUnit report but the output directory can't be written, you would expect the run to exit with an error no ?

aurelien-reeves: I agree with @vincent-psarga 
People who expect reports to be published from a CI for example should be noticed that something went wrong.

I run a few tests with cucumber-ruby. I was expecting a non-zero exit code, it was not the case. I think I'll open an issue in cucumber-ruby about that.

I will try the same with cucumber-jvm

davidjgoss: Makes sense to me. I think it’s fine to kick the option down the road and see if there’s demand.

PR #1613, opened 2021-03-16
Removed wrong documentation

silversonicaxel:


silversonicaxel:
this is the follow up of this one https://github.com/cucumber/cucumber-js/pull/1611

PR #1614, opened 2021-03-16
Fix inconsistent tests

aurelien-reeves:
Tests which uses the FakeReportServer seems to be inconsistent.

In that PR, one test in http_stream_spec tries to reproduce the error


aurelien-reeves:
Will reopen it with a cleaner history

PR #1617, opened 2021-03-19
Error in CI with HttpStream/FakeReportServer

aurelien-reeves:
Tests are not stable enough on the CI
It seems to be due to a race condition around http_stream and/or the fakeReportServer used for the tests.


mattwynne:
@aurelien-reeves is this still a live issue?

aurelien-reeves:
Yes, the issue is still live
But it happens far less often that it used to.

We may close that one until it raises often again

mattwynne:
OK I think we have identified the culprit and have a fix.

Pairing with @aurelien-reeves we isolated the problem down to starting the `FakeReportServer`: when the problem occurs, we get a `listening` event back from the HTTP server, but in fact the server isn't responding to connection attempts.

On a hunch, I tried using a fixed port instead of this trick of passing a `0` port to `server.listen` (which tells the server you want it to find a free port) and it started working.

In order to avoid hard-coding a port, I've used a library called `portfinder` to look up a free port to use, instead of letting the server do that. It seems to be more reliable.

Once I had it working I dismantled some complexity in the `FakeReportServer` around managing/closing sockets, which doesn't seem to be necessary.

mattwynne:
One thing that occurs to me is that we could move all of the cucumber-reports integration stuff out into some kind of plugin. It might be a useful exercise to create the extension points, and it would also mean we could leave these tests (and code) out of our main codebase / CI run.

aslakhellesoy:
Nice!!!!

davidjgoss (approved):

mattwynne:
Well, I guess I spoke too soon! https://github.com/cucumber/cucumber-js/runs/5310937749?check_suite_focus=true

Maybe we need a new test to isolate this issue. That previous one definitely seems to be passing OK now.

aurelien-reeves:
Actually that "ERR_PREMATURE_CLOSE" error is the one we originally had and tried to reproduce and fix.

mattwynne:
> Actually that "ERR_PREMATURE_CLOSE" error is the one we originally had and tried to reproduce and fix.

Yeah, it's the same one I witnessed [here](https://github.com/cucumber/cucumber-js/pull/1617#issuecomment-1035328082), too.

Interesting that it happened on Windows both times. I wonder if that's relevant.

mattwynne:
Seems relevant: https://stackoverflow.com/questions/55959479/error-err-stream-premature-close-premature-close-in-node-pipeline-stream

mattwynne:
I would like us to still merge the changes in this PR as they're an improvement, even if they don't fix *all* the flickers in this area of the test suite.

PR #1637, opened 2021-04-02
refresh cucumber core dependencies

davidjgoss:
And:

- Use the new `@gherkin-streams` package where needed
- Add feature coverage for tags on rules
- Tell renovate to batch together core dependency updates in future


yarn.lock
```diff
@@ -229,7 +238,24 @@
...
+"@cucumber/messages@15.0.0", "@cucumber/messages@^15.0.0":
+  version "15.0.0"
+  resolved "https://registry.yarnpkg.com/@cucumber/messages/-/messages-15.0.0.tgz#58e6541a6c21c4f0d3d4a82c24bbbfe72a6ff94c"
+  integrity sha512-LtxzSCRmYZTAKO6ucAcMflz0u90l2fev539OG+EioJ26F14KmmtxZwGabfjTxLf8NgyKeWsO8TGI2G3z4Kjr+A==
+  dependencies:
+    "@types/uuid" "^8.3.0"
+    protobufjs "^6.10.2"
+    uuid "^8.3.2"
+
+"@cucumber/messages@^14.0.1":
```
davidjgoss, line 258: This version still being pulled in by `html-formatter`. Shouldn't be an issue but would be good to resolve (by making a release of `html-formatter`).

davidjgoss:
A couple of issues here I might need a hand with @aslakhellesoy @aurelien-reeves:

- `@cucumber/compatibility-kit@4.0.0` is [tagged](https://github.com/cucumber/cucumber/releases/tag/compatibility-kit%2Fv4.0.0) in the monorepo but hasn't been [published to npm](https://www.npmjs.com/package/@cucumber/compatibility-kit)
- `@cucumber/query@9.0.0` is published to npm but the package is missing some key files, see https://cucumberbdd.slack.com/archives/C62D0FK0E/p1617374004111900

yarn.lock
mpkorstanje: I'm stuck on that too. Looks like some issues with the build have to be fixed first.

https://github.com/cucumber/cucumber/issues/1448
https://github.com/cucumber/cucumber-jvm/pull/2268

aslakhellesoy: The release of `@cucumber/react 13.0.0` and  `@cucumber/html-formatter 13.0.0` should align all versions. I suggest we wait to merge this until those libs are released and we've upgraded this PR to use them.

aslakhellesoy (changes requested):

yarn.lock
davidjgoss: Agreed

davidjgoss: And updated!

PR #1650, opened 2021-04-22
cli: undeprecate json formatter

### LINKED ISSUE #1250: Remove JSON formatter
As described in #1249 the plan is to deprecate the `json` formatter in version 7. In version 8 we should remove it.

After the removal people can still generate JSON reports with the standalone [json formatter](https://github.com/cucumber/cucumber/tree/master/json-formatter). There are prebuilt [executables](https://github.com/cucumber/cucumber/releases/tag/json-formatter%2Fv2.0.0) that will convert output from the `protobuf` formatter to JSON. This is intended to be used only by people who have to use tools that still only support the old JSON format.

davidjgoss:
Closes #1250 as agreed in today's community call for all implementations.

The previous deprecation has resulted in some hesitation to upgrade from users who use tools that rely on the JSON formatter output.

The tone now is about "maintenance mode" - it'll stay functional, but won't benefit from any new features. As the documentation and tooling around messages improves, we'll gradually encourage authors of maintained tools to switch over to that. Feedback welcome on the exact wording. Also, the noisy console warning is gone.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -59,7 +59,7 @@ If multiple formats are specified with the same output, only the last is used.
 
 * **message** - prints each [message](https://github.com/cucumber/cucumber/tree/master/cucumber-messages) in NDJSON form, which can then be consumed by other tools.
 * **html** - prints a rich HTML report to a standalone page
-* **json** - prints the feature as JSON. *Note: this formatter is deprecated and will be removed in the next major release. Where you need a structured data representation of your test run, it's best to use the `message` formatter. For legacy tools that depend on the deprecated JSON format, a standalone formatter is available (see https://github.com/cucumber/cucumber/tree/master/json-formatter).
+* **json** - prints the feature as JSON. *Note: this formatter is in maintenance mode and won't have new features added to it. Where you need a structured data representation of your test run, it's best to use the `message` formatter. Tools that rely on this formatter will continue to work, but are encouraged to migrate to consume the `message` output instead.*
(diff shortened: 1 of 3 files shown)


docs/cli.md
```diff
@@ -59,7 +59,7 @@ If multiple formats are specified with the same output, only the last is used.
 
 * **message** - prints each [message](https://github.com/cucumber/cucumber/tree/master/cucumber-messages) in NDJSON form, which can then be consumed by other tools.
 * **html** - prints a rich HTML report to a standalone page
-* **json** - prints the feature as JSON. *Note: this formatter is deprecated and will be removed in the next major release. Where you need a structured data representation of your test run, it's best to use the `message` formatter. For legacy tools that depend on the deprecated JSON format, a standalone formatter is available (see https://github.com/cucumber/cucumber/tree/master/json-formatter).
+* **json** - prints the feature as JSON. *Note: this formatter is in maintenance mode and won't have new features added to it. Where you need a structured data representation of your test run, it's best to use the `message` formatter. Tools that rely on this formatter will continue to work, but are encouraged to migrate to consume the `message` output instead.*
```
aurelien-reeves, line 62: Why the `*` at the end of the sentence?

davidjgoss: There’s one before “Note”, to italicise that section

aurelien-reeves: 👍

I looked for it, did not found it before. Now I do 😅

mattwynne (approved):

Issue #1532, opened 2021-01-02
compilation issues with typescript


davidjgoss:
When compiling a TypeScript project using v7 with this pretty standard `tsconfig.json`:

```jsom
{
  "include": [
    "support/**/*.ts"
  ],
  "compilerOptions": {
    "target": "es2017",
    "module": "commonjs",
    "moduleResolution": "node",
    "noEmit": true,
    "noImplicitAny": true,
    "noImplicitReturns": true
  }
}
```

We see these two compilation errors:

```
node_modules/@cucumber/cucumber/lib/runtime/attachment_manager/index.d.ts:2:8 - error TS1259: Module '"stream"' can only be default-imported using the 'esModuleInterop' flag

2 import stream from 'stream';
         ~~~~~~

  node_modules/@types/node/stream.d.ts:331:5
    331     export = internal;
            ~~~~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

node_modules/@cucumber/messages/dist/src/index.d.ts:9:25 - error TS2732: Cannot find module '../package.json'. Consider using '--resolveJsonModule' to import module with '.json' extension

9 import { version } from '../package.json';
                          ~~~~~~~~~~~~~~~~~
```

This can be easily worked around by adding `esModuleInterop` and `resolveJsonModule` compiler options, but it'd be better if we didn't have to.

- [x] Remove need for `esModuleInterop`
- [ ] Remove need for `resolveJsonModule`
- [ ] Remove need for `@types/lodash`

jan-molak:
@davidjgoss I agree, this is problematic because any project that depends on `@cucumber/cucumber` will now have to enable those two flags. I'd prefer to avoid having to do it, if possible.

`esModuleInterop` could be avoided by using:
```typescript
import * as stream from 'stream'
```
instead of:
```typescript
import stream from 'stream'
```

And since the only reason to use `resolveJsonModule` seems to be to read the version of the `@cucumber/cucumber` module, perhaps `require` or even `fs.readFileSync` could be used instead?

jan-molak:
@davidjgoss - I've just noticed the same issue with `resolveJsonModule` present in `@cucumber/cucumber-messages`:

```
node_modules/@cucumber/messages/dist/src/index.d.ts:9:25 - error TS2732: Cannot find module '../package.json'. Consider using '--resolveJsonModule' to import module with '.json' extension.
```

[...]

PR #1662, opened 2021-05-02
replace use of Dictionary with Record

davidjgoss:
Helps with #1532 - avoids an implicit dependency on `@types/lodash` where TypeScript users reference some of our models and functions.

### DIFF
--- a/src/models/data_table.ts
+++ b/src/models/data_table.ts
@@ -33,7 +33,7 @@ export default class DataTable {
     return copy
   }
 
-  rowsHash(): Record<string, any> {
+  rowsHash(): Record<string, string> {
     const rows = this.raw()
     const everyRowHasTwoColumns = _.every(rows, (row) => row.length === 2)
     if (!everyRowHasTwoColumns) {
--- a/src/models/data_table.ts
+++ b/src/models/data_table.ts
@@ -1,4 +1,4 @@
-import _, { Dictionary } from 'lodash'
+import _ from 'lodash'
 import { messages } from '@cucumber/messages'
 
 export default class DataTable {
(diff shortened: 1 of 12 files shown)


src/models/data_table.ts
```diff
@@ -33,7 +33,7 @@ export default class DataTable {
     return copy
   }
 
-  rowsHash(): Dictionary<any> {
+  rowsHash(): Record<string, any> {
```
aurelien-reeves, line 36: As `this.raw()` returns a `string[][]`, would it be possible for rowsHash to return a `Record<string, string>` instead of `Record<string, any>`?

aurelien-reeves (commented):
`src/formatter/json_formatter.ts:316`: could we replace `{ [id: string]: IScenario }` with a `Record`?

src/models/data_table.ts
davidjgoss: You're right, good catch - will change.

davidjgoss:
@aurelien-reeves thanks for feedback - addressed now

aurelien-reeves (approved):

aurelien-reeves:
@davidjgoss LGTM

Quick question - off topic: I see that cucumber-js uses a lot lodash even when it seems it could be replaced with native functions. Is there a reason for that? Are we encourage to use lodash? To try to remove it? To let it as-is?

davidjgoss:
@aurelien-reeves good point. I think a lot of the lodash usage is valid, but also probably a lot of stuff like `map`, `each` and `filter` predates the native functions being available in Node. If you raise a ticket we can maybe chip away at those?

aurelien-reeves:
@davidjgoss actually I was more thinking about taking care of that as part of other work. For example my last PR, I let some lodash usage which I could have replaced by native methods if I had been aware I could.

Would you prefer a dedicated PR?

davidjgoss:
@aurelien-reeves ah no worries - no I think it's fine to do them piecemeal as you touch areas.

PR #1694, opened 2021-06-04
Update migration links to messages specifications

aurelien-reeves:
Refs. https://github.com/cucumber/cucumber-js/pull/1692#discussion_r645274491

Use links to tags instead of links to main branch to avoid the links being dead in a few years.

I am still puzzled about the maintenance of those links. But as it is about the migration from 6 to 7, we can argue that it does not have to be updated for each new version of messages.


davidjgoss:
> But as it is about the migration from 6 to 7, we can argue that it does not have to be updated for each new version of messages.

Yeah I think that's fair. I was hesitant to link to a tag for fear of the tag then being outdated, but as you say the migration guide should become defunct anyway.

Side note: it would be cool if we had nice documentation of the message schema on a website somewhere we could link to...

davidjgoss (approved):

aurelien-reeves:
> Side note: it would be cool if we had nice documentation of the message schema on a website somewhere we could link to...

With proto, we generated a messages.md which was nicer to look at than the proto file: https://github.com/cucumber/common/blob/main/messages/messages.md

We have to keep doing that with json schema. For now it is still up-to-date. But it won't be if we do not find a new way to keep it up-to-date with json schemas.

PR #1697, opened 2021-06-11
debt: add things to main entry point that people need

davidjgoss:
Much of the list from https://github.com/cucumber/cucumber-js/issues/1646#issue-863352190. Most of the rest already available under formatter helpers, probably just need some manual editing in those projects.

### DIFF
--- a/src/support_code_library_builder/types.ts
+++ b/src/support_code_library_builder/types.ts
@@ -25,12 +25,12 @@ export interface ITestStepHookParameter {
 
 export type TestCaseHookFunction<WorldType> = (
   this: WorldType,
-  arg?: ITestCaseHookParameter
+  arg: ITestCaseHookParameter
 ) => any | Promise<any>
 
 export type TestStepHookFunction<WorldType> = (
   this: WorldType,
-  arg?: ITestStepHookParameter
+  arg: ITestStepHookParameter
 ) => void
 
 export type TestStepFunction<WorldType> = (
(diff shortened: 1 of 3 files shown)


src/support_code_library_builder/types.ts
```diff
@@ -25,12 +25,12 @@ export interface ITestStepHookParameter {
 
 export type TestCaseHookFunction<WorldType> = (
   this: WorldType,
-  arg?: ITestCaseHookParameter
+  arg: ITestCaseHookParameter
```
davidjgoss, line 28: This was a mistake from #1690. The argument isn't optional - we always pass it in - it's just up to the user-authored code whether it does anything with it.

aurelien-reeves:
Looks good.

I wonder: is there a way to reproduce through a test the issue reported on #1646?

davidjgoss:
>  wonder: is there a way to reproduce through a test the issue reported on #1646?

Yeah, I added a feature test in https://github.com/cucumber/cucumber-js/pull/1649 that hits the error when trying to do deep imports after the ESM change. Still working on if/how to handle that.

aurelien-reeves:
> 
> 
> > wonder: is there a way to reproduce through a test the issue reported on #1646?
> 
> Yeah, I added a feature test in #1649 that hits the error when trying to do deep imports after the ESM change. Still working on if/how to handle that.

Great!

aurelien-reeves (approved):

PR #1705, opened 2021-06-25
Add more arrow function warnings

aslakhellesoy:
# Description

Some users use arrow functions without realising that they cannot access the world from them.

# Motivation & context

This is currently described in the FAQ, but people still miss it. I've therefore added more sign-posts from the hook, step definition and parameter type docs.

I also tweaked the FAQ entry a little.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -2,12 +2,15 @@
 
 ## The world instance isn’t available in my hooks or step definitions.
 
-This has frequently been caused by the use of ES6 arrow functions.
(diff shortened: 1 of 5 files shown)


davidjgoss (approved):

davidjgoss:
I wondered if we could detect this at runtime and emit a warning but seems it's [not straightforward](https://stackoverflow.com/questions/28222228/javascript-es6-test-for-arrow-function-built-in-function-regular-function) :(

PR #1730, opened 2021-07-14
Explain how to use yarn to list commands

cucumber-ensemble:
Co-authored-by: Matt Wynne <matt@mattwynne.net>


mattwynne:
This is just a little test of a collaborative mobbing environment I've been setting up.

CONTRIBUTING.md
```diff
@@ -9,7 +9,7 @@ Before anything else, thank you. Thank you for taking some of your precious time
 
 ## Tests
 
-See the `package.json` scripts section for how to run each category of tests.
+Now type `yarn run` or see the `package.json` scripts section for how to run each category of tests.
```
aslakhellesoy, line 12: I'd prefer if we didn't recommend the use of `yarn` and use `npm` instead. Because it's the standard.

aslakhellesoy (dismissed):

CONTRIBUTING.md
davidjgoss: I think the `yarn` usage is correct currently because the project is developed with yarn (the `yarn.lock` file, CI builds). That said I would be fine with changing to npm to be more standard.

aurelien-reeves: `yarn` is a pretty popular alternative to `npm`. I would go for documenting usage of both, or none. Eventually only `npm`. But definitely not `yarn` only.

Here, listing available scripts with npm is possible with `npm run` as well as with yarn

mattwynne: Right now, the project uses yarn. If we want to switch to npm that's fine, but it's a separate issue to improving the current docs.

PR #1765, opened 2021-08-09
Add retro notes from new contributors ensemble

mattwynne:
Each of the new-contributors ensemble sessions we've run, we've held a 15-minute
retrospective at the end.

I'm not sure where's the best place to put these notes, but I felt like maybe it didn't really do any harm just to put them right here. I'm hopeful we can start to capture more feedback about specific changes we can make to the codebase to make it more welcoming, as time goes on.


doc/retro/2021/08/06.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Actions / decisions
+
+* Next time, continue with fixing [#1136]
+
+[@artismarti]: https://github.com/artismarti
+[@mattwynne]: https://github.com/mattwynne
+[@16sheep]: https://github.com/16sheep
+[#1764]: https://github.com/cucumber/cucumber-js/pull/1764
+[#1136]: https://github.com/cucumber/cucumber-js/issues/1163
```
aslakhellesoy, line 37: 1136 or 1163?

davidjgoss (commented):
Looks great, thanks for taking the time to write up.

I'd just nitpick on the dir path `doc/` - we already have `docs/`, so I'd maybe suggest putting under there, or maybe under a new path like `community/`?

doc/retro/2021/08/06.md
mattwynne: Thanks!

mattwynne:
Thanks @davidjgoss I hadn't spotted that. I've moved it into `/docs`

PR #1770, opened 2021-08-17
New Feature: shared context in beforeAll and afterAll hooks

### LINKED ISSUE #1393: Feature Request: Give BeforeAll a reference to the Cucumber World object
To get around this we've been rolling our own DI which is cumbersome. If this is by intention, I'm wondering why BeforeAll is not given this reference?

BlueMona:
# Description

This is a draft failing scenario to try and describe what @charlierudolph and
@davidjgoss are describing in https://github.com/cucumber/cucumber-js/issues/1393

I'd like to work on a fix but thought it would be best to get feedback on the scenario first.

# Motivation & context

Fixes #1393

## Type of change

# Checklist:


mattwynne:
The scenario we have so far doesn't address the issue around immutability. I suggest we add another scenario showing that even if the first scenario mutates the value that was set in the `BeforeAll` hook, the second scenario gets the original value again. (i.e. we can't use it to leak state between scenarios; each new World instance gets a fresh copy of the `testRunContext`)

davidjgoss:
@BlueMona thanks for getting this going, and great idea to start with the scenario. I think my only suggestion would be for the stuff that we set in `BeforeAll` to be set on the world under its own `testRunContext` property rather than mixed in, so like:

```
Given('first step', function() {
  expect(this.testRunContext.myVar.foo).to.eql(1) 
})
```

Then it's like a sibling of `parameters` and `attach` that you can access from your world (including a custom one).

@mattwynne seems a good direction re immutability.

BlueMona:
The scenario is now passing, but there are a few questions / things left to do. 
- is object an ok type for testRunContext? 
- testRunContext should likely be a property of world, instead of being merged into the root of the object, which would help avoid potential conflicts
- how do we implement this for the worker and how is it tested? 
- do we need a unit test for the testCaseRunner? 
- what about immutability? Should we add a scenario testing for it?

davidjgoss:
Thanks for the update @BlueMona, this is looking good. To your points:

> is object an ok type for testRunContext?

It's kinda the same need as `worldParameters`, which we type as `any`. Technically it should be a JSON-serialisable object so `any` is a bit broad, but there's no nice way to express that in TypeScript at the moment, so I'd say `any` is good for now.

> testRunContext should likely be a property of world, instead of being merged into the root of the object, which would help avoid potential conflicts

I definitely agree, and I'd even go as far as giving each `World` instance a clone of the original `testRunContext` so that if users do modify it, the changes won't bleed into other scenarios running.

> how do we implement this for the worker and how is it tested?

Good question, I'll get back to you! We run the test run hooks once for each worker which makes things...interesting. Testing of parallel stuff is generally in the feature tests.

> what about immutability? Should we add a scenario testing for it?

Yep would be good to test this. Hopefully the cloning idea above would have the desired effect of (enough) immutability. Marking it `readonly` in TypeScript would be an easy win too.

mattwynne:
@davidjgoss thanks for the quick feedback!

I'm wondering if we could achieve the immutability by copying the `testRunContext` when we create the merged `this` object in `invokeStep`.

i.e. something like 

    const world = Object.assign(this.world, { testRunContext: { ...this.testRunContext }})

WDYT?

davidjgoss:
@mattwynne yeah that's basically what I was suggesting, although we'd want a [deep clone](https://www.npmjs.com/package/lodash.clonedeep) since doing a spread just gives you a shallow clone.

mattwynne:
@BlueMona I've extracted a function to run the `BeforeAll` hooks in #1933. We'll want to merge that into this PR before proceeding.

One slightly tricky part of that merge will be sharing the `thisArg` / `testRunContext` with the new function. I suggest adding it as a new argument to the function, but you could also consider making it be the return value, I suppose.

kamelbou:
Hello,

Is it plan to merge this great feature soon ? :)

Thank you

mattwynne:
@BlueMona I had a little look after we finished our session, and I think we still have a couple of other places we need to make a similar change, so that the `thisArg` object in the `runTestRunHooks` function has a _property_ of `testRunContext` rather than _being_ the `testRunContext`.

@davidjgoss as we start to make this distinction between the two, I wonder about how typing could help us here. We could define an explicit `type TestRunContext = any` type and use that everywhere we expect to get a `TestRunContext`, but what about the `world` / `thisArg`? Could we type that as `type World = { testRunContext: TestRunContext }` too? What would be the implications for users who have a custom world? Would this end up being a breaking change?

davidjgoss:
This is shaping up nicely.

@mattwynne for the world I think we should use the existing structures i.e. add `testRunContext` to the constructor object for the built in world and its interface ([here](https://github.com/cucumber/cucumber-js/blob/main/src/support_code_library_builder/world.ts)) and then pass it in when we create the world instance ([here](https://github.com/cucumber/cucumber-js/blob/main/src/runtime/test_case_runner.ts#L92)) - that way we don't have to worry about doing things at step level. I think it'd make sense to type it as `readonly` (like the other props on that object) and also perhaps to [deep clone](https://www.npmjs.com/package/lodash.clonedeep) it each time we do that.

davidjgoss:
Also there are some tests specifically around types that you might want to look at and/or extend https://github.com/cucumber/cucumber-js/blob/main/test-d/world.ts

Wepstern:
Hi,

I am very happy to have found this ticket. It would give us a very powerful tool to describe and manage our test environment. This would avoid the test environment management scattered throughout the code that can cause a lot of headaches.

Huge high five for that! :)

davidjgoss (changes requested):
Looking fantastic, thanks for taking this on Mona. And sorry for the delayed review.

One thing I think we're missing is the correct typing of the `BeforeAll` and `AfterAll` functions so people can access `this.testRunContext` while using TypeScript without a compilation error.

If you go to `test-d/hooks.ts` and add:

```ts
// should allow test run context via this arg in test run hooks
BeforeAll(async function () {
  this.testRunContext.foo = 'bar'
})
AfterAll(async function () {
  this.testRunContext
})
```

Then that will currently fail if you run `npm run types-test`. I think the place to fix it is in `src/support_code_library_builder/types.ts`. We have explicit types for `TestCaseHookFunction` etc there and so it would be good to follow that pattern - although we don't need the complexity with the generics for world there, so it should be easier, something like:

```ts
export type TestRunHookFunction = (
  this: { testRunContext: TestRunContext }
) => any | Promise<any>
```

We'll also need to update some docs:
- https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/api_reference.md#afteralloptions-fn - basic API doc for the functions
- https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/hooks.md#beforeall--afterall - explain how to set test run context
- https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/world.md - explain you can access test run context read only

I'm very happy to help out with any of this as required, just let me know. Thanks again!

ksathyanm:
Just curious if we have any updates or plan for this PR?

mattwynne:
@BlueMona ping me if you want to pair on finishing this off. I'm here to help if you want it.

BlueMona:
Let’s do it this week then!

mattwynne:
> Let’s do it this week then!
> […](#)

@BlueMona great!

Can you use https://calendly.com/mattwynne to book a slot?

LANDG-MartinP:
any news on this PR?

davidjgoss:
Just noting here the idea raised on the issue of, rather than adding a new `testRunContext` concept, just passing in the world parameters and allowing a `BeforeAll` to return a fresh world parameters if it wants to. So like:

```
BeforeAll(async function() {
  this.token = await authenticate(this.parameters.url, this.parameters.clientId, this.parameters.clientSecret)
  return {
    ...this.parameters,
    token
  }
})
```

I think this would solve all the use cases nicely and this PR could be adapted to it without much work - @BlueMona @mattwynne I'm happy to pick this up if you're time-constrained, since I'm working in this area already.

Also somewhat related is #2356 which I'm looking at today.

BlueMona:
David,

Yes, please do.

davidjgoss (approved):
After some conflict untangling, this fitted together nicely with some other recent changes.

Many thanks @BlueMona for doing most of this!

aslakhellesoy:
Hi @BlueMona,

Aslak Hellesøy
Creator of Cucumber

iamkenos:
This one's a great holiday gift. Thanks team 🖤

mattwynne:
🎉 nice one @BlueMona!

PR #1775, opened 2021-08-21
feat: add pickleStep to step hook function arg

### LINKED ISSUE #1740: Unable to get StepName from ITestStepHookParameter in BeforeStep hook
**Describe the bug**
I am trying to pull current step name out of ITestStepHookParameter in the BeforeStep hook:

```
BeforeStep(async function(testStep) {
     // Should be step name, is scenario name instead
     const stepName = testStep.pickle.name;

    // stepObj is undefined
    const stepObj = testStep.pickle.steps.filter(step => {
        return step.id === testStep.testStepId
    });
});
```

**Cucumber-js version: 7.3.1**

**Expected behavior**
I would expect the `testStep.testStepId` to be present in the array of steps in the `testStep.pickle.steps` object. However, that's not the case and the `testStepId` does not much any of the steps in the array.

Also, the `testStep.pickle.name` contains the name of the whole scenario, not the concrete step. 
As a result, it is not able to pull out the step name inside the BeforeStep hook.

davidjgoss:
# Description

`BeforeStep` and `AfterStep` hook functions now have access to the `pickleStep` in their argument object.

# Motivation & context

Step-level hook functions currently have access to the `pickle`, but don't have a way of knowing which step in the pickle is being run. I think this was just an oversight when we implemented step hooks where the argument object was based on the same one for scenario hooks.

Fixes #1740.

## Type of change

# Checklist:


aurelien-reeves (approved):
Looks good to me 👌

PR #1793, opened 2021-09-01
Increase precision of test case duration measurements.

joac:
# Description
I was working on a[ custom formatter for `pickled-cucumber` ](https://github.com/muralco/pickled-cucumber/blob/master/src/formatter/profile-jsonl.ts) and noticed that fast tests always reported 0 as the duration, after some debugging I noticed that we were not using the most accurate clock, so I changed it for `performance.now` from perf hooks
https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_now
As a side effect, that seems to be a monotonic clock, making the
measurements not affected by leap seconds

# Motivation & context

The durations numbers reported by cucumber were not accurate.

## Type of change

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,10 @@ See the [migration guide](./docs/migration.md) for details of how to migrate fro
 
 ### Added
 
+* `--config` option to the CLI. It allows you to specify a configuration file other than `cucumber.js`.
+  See [docs/profiles.md](./docs/profiles.md#using-another-file-than-cucumberjs) for more info.
+  [#1794](https://github.com/cucumber/cucumber-js/pull/1794)
+
 ### Changed
 * Use performance timers for test case duration measurement.
   [#1793](https://github.com/cucumber/cucumber-js/pull/1793)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ See the [migration guide](./docs/migration.md) for details of how to migrate fro
 ### Added
 
 ### Changed
+* Use performance timers for test case duration measurement.
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ See the [migration guide](./docs/migration.md) for details of how to migrate fro
 
 ### Changed
 * Use performance timers for test case duration measurement.
+  [#1793](https://github.com/cucumber/cucumber-js/pull/1793)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ See the [migration guide](./docs/migration.md) for details of how to migrate fro
 
 ### Added
 
+* `BeforeStep` and `AfterStep` hook functions now have access to the `pickleStep` in their argument object.
 * `--config` option to the CLI. It allows you to specify a configuration file other than `cucumber.js`.
   See [docs/profiles.md](./docs/profiles.md#using-another-file-than-cucumberjs) for more info.
   [#1794](https://github.com/cucumber/cucumber-js/pull/1794)
--- a/features/usage_formatter.feature
+++ b/features/usage_formatter.feature
@@ -33,21 +33,21 @@ Feature: usage formatter
     When I run cucumber-js with `--format usage`
     Then it outputs the text:
       """
-      ┌────────────────────┬──────────┬───────────────────────────────────────┐
-      │ Pattern / Text     │ Duration │ Location                              │
-      ├────────────────────┼──────────┼───────────────────────────────────────┤
-      │ step A             │ <d>ms    │ features/step_definitions/steps.js:3  │
-      │   step A           │ <d>ms    │ features/a.feature:3                  │
-      ├────────────────────┼──────────┼───────────────────────────────────────┤
-      │ /^(slow )?step B$/ │ <d>ms    │ features/step_definitions/steps.js:4  │
-      │   slow step B      │ <d>ms    │ features/a.feature:5                  │
-      │   step B           │ <d>ms    │ features/a.feature:4                  │
-      ├────────────────────┼──────────┼───────────────────────────────────────┤
-      │ step C             │ <d>ms    │ features/step_definitions/steps.js:11 │
-      │   step C           │ <d>ms    │ features/a.feature:6                  │
-      ├────────────────────┼──────────┼───────────────────────────────────────┤
-      │ step D             │ UNUSED   │ features/step_definitions/steps.js:12 │
-      └────────────────────┴──────────┴───────────────────────────────────────┘
+      ┌────────────────────┬──────────────┬───────────────────────────────────────┐
+      │ Pattern / Text     │ Duration     │ Location                              │
+      ├────────────────────┼──────────────┼───────────────────────────────────────┤
+      │ step A             │ <d>ms        │ features/step_definitions/steps.js:3  │
+      │   step A           │ <d>ms        │ features/a.feature:3                  │
+      ├────────────────────┼──────────────┼───────────────────────────────────────┤
+      │ /^(slow )?step B$/ │ <d>ms        │ features/step_definitions/steps.js:4  │
+      │   slow step B      │ <d>ms        │ features/a.feature:5                  │
+      │   step B           │ <d>ms        │ features/a.feature:4                  │
+      ├────────────────────┼──────────────┼───────────────────────────────────────┤
+      │ step C             │ <d>ms        │ features/step_definitions/steps.js:11 │
+      │   step C           │ <d>ms        │ features/a.feature:6                  │
+      ├────────────────────┼──────────────┼───────────────────────────────────────┤
+      │ step D             │ UNUSED       │ features/step_definitions/steps.js:12 │
+      └────────────────────┴──────────────┴───────────────────────────────────────┘
       """
 
   Scenario: only list 5 slowest matches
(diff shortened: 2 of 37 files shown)


joac:
Test are failing due to inaccurate `nanos`, It will be really hard to fix those tests, due to the jitter we can expect between test runs. I will need some help here. :pray:

aurelien-reeves:
For the tests, we are using https://github.com/sinonjs/fake-timers to have predictable timestamps.
sinonjs/fake-timers can fake performance.now, but only in browser. So we may need to find an alternative.

aurelien-reeves:
Found how to fix the tests: this is the why of the "methods" object.

Add the "performance" object to the "methods":

```typescript

const methods: any = {
  // snip...
  Date,
  performance,
  // snip...
}
```

And use it in `getTimestamp`:
```typescript
function getTimestamp(): number {
  return methods.performance.now()
}
```

In the unit tests, we are installing sinonjs/fake-timers with that "methods" object. Adding the "performance" entry in it, sinonjs/fake-timer will found it and replace it with its fake one

joac:
@aurelien-reeves will research a bit about alternatives to fix that, thanks!

aurelien-reeves:
@joac I've found a way, please check my previous message

joac:
I had to increase the colum size for durations in msecs on usage formatter tests, due to bigger numbers being reported, do you think is ok @aurelien-reeves o we should use `Math.round` or `Math.floor ` on the formatter itself?

joac:
Seems that the tests that break are not using the fake timers. I don't understand well the initialization process of the test suite :cry:

aurelien-reeves:
> Seems that the tests that break are not using the fake timers. I don't understand well the initialization process of the test suite 😢

You do not have to force push a single commit each time.
Actually that prevent us to see the history of the CI.

> 
> 
> I had to increase the colum size for durations in msecs on usage formatter tests, due to bigger numbers being reported, do you think is ok @aurelien-reeves o we should use `Math.round` or `Math.floor ` on the formatter itself?

I think that technically, it looks good to me.
But I would wait for the opinion of @davidjgoss, @charlierudolph or @aslakhellesoy about your feature.

aurelien-reeves:
I think you should revert your changes on ` features/support/hooks.ts`
It seems to be responsible for the CI being stuck

features/usage_formatter.feature
```diff
@@ -81,15 +81,15 @@ Feature: usage formatter
...
-      ├──────────────────┼──────────┼──────────────────────────────────────┤
-      │ /^(slow )?step$/ │ <d>ms    │ features/step_definitions/steps.js:3 │
-      │   slow step      │ <d>ms    │ features/a.feature:3                 │
-      │   slow step      │ <d>ms    │ features/a.feature:3                 │
-      │   slow step      │ <d>ms    │ features/a.feature:3                 │
-      │   slow step      │ <d>ms    │ features/a.feature:3                 │
-      │   slow step      │ <d>ms    │ features/a.feature:3                 │
-      │   5 more         │          │                                      │
-      └──────────────────┴──────────┴──────────────────────────────────────┘
+      ┌──────────────────┬──────────────┬──────────────────────────────────────┐
```
aurelien-reeves, line 84: I think you are just missing an extra `─` for the `duration` column

joac: It  works some times on linux and fails on windows, I'm thinking about a better solution, probably mocking the time on said tests.

joac: I just made the normalization function take care of making the number of dashes not matter.
Linux and windows were having different behaviors.

joac:
> I think you should revert your changes on ` features/support/hooks.ts`
> It seems to be responsible for the CI being stuck
Was a failed attempt to fake the timers (a bad import). I changed the approach for testing and CI is green now.

aurelien-reeves (commented):
Nice, well done!

It looks good to me. I like your PR!

But as I said, I prefer to wait for some more experienced folks to review it before approving it. I don't know if using the monotonic `performance` clock could have some undesirable impact somewhere with the reports.

@davidjgoss @charlierudolph @aslakhellesoy any objection with this?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ See the [migration guide](./docs/migration.md) for details of how to migrate fro
 ### Added
 
 ### Changed
+* Use performance timers for test case duration measurement.
```
aurelien-reeves, line 23: Could you please add a link to the current PR as it is done elsewhere in the CHANGELOG please?

src/time.ts
```diff
@@ -22,7 +24,7 @@ if (typeof setImmediate !== 'undefined') {
 }
 
 function getTimestamp(): number {
```
davidjgoss, line 26: So this used to return millis since epoc and now millis since process started - which I think is fine because we only use it for duration, not a report an actual point in time, but I'll double check that.

aurelien-reeves: Yes, that's it.
Also, that is not an integer anymore, but a float.

aslakhellesoy: > Also, that is not an integer anymore, but a float.

@aurelien-reeves what do you mean? JavaScript doesn't have an integer or float type, only `number`

aslakhellesoy:
> But as I said, I prefer to wait for some more experienced folks to review it before approving it. I don't know if using the monotonic performance clock could have some undesirable impact somewhere with the reports.

As long as the measurement unit is the same (milliseconds I presume) I have no objections. Using a monotonic clock is the correct way to measure duration in any software system (using a wall clock is always wrong).

src/time.ts
aurelien-reeves: I mean, before it looked like integers, ie 500ms. With the proposed change, it will look like double, ie 500.11125473917ms.

It may puzzle some users

aslakhellesoy: I see what you mean. The reports format the durations as “human friendly” so I don’t think it will make a difference.

aurelien-reeves (approved):

davidjgoss:
> I had to increase the colum size for durations in msecs on usage formatter tests

Do we have an example of how this comes out?

aurelien-reeves:
> 
> 
> > I had to increase the colum size for durations in msecs on usage formatter tests
> 
> Do we have an example of how this comes out?

Looks like that:

```console
$ node ./bin/cucumber-js --format usage
┌────────────────────────────────────────────────────────────────────────────────────────────────────────────────┬───────────────┬───────────────────────────────────────────────────────┐
│ Pattern / Text                                                                                                 │ Duration      │ Location                                              │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────────────────────────────────────────────┤
│ /^I run cucumber-js(?: with `(|.+)`)?$/                                                                        │ 47788.62ms    │ features/step_definitions/cli_steps.ts:19             │
│   I run cucumber-js                                                                                            │ 11874726.13ms │ features/fake_time.feature:39                         │
│   I run cucumber-js                                                                                            │ 1445.457999ms │ features/exit.feature:28                              │
│   I run cucumber-js with `--parallel 2`                                                                        │ 1253.9458ms   │ features/parallel.feature:21                          │
│   I run cucumber-js                                                                                            │ 1014.318199ms │ features/before_after_all_hook_timeouts.feature:51    │
│   I run cucumber-js                                                                                            │ 1009.0063ms   │ features/step_definition_timeouts.feature:105         │
│   244 more                                                                                                     │               │                                                       │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────────────────────────────────────────────┤
│ /^I run cucumber-js with env `(|.+)`$/                                                                         │ 416.24ms      │ features/step_definitions/cli_steps.ts:40             │
│   I run cucumber-js with env `CUCUMBER_PUBLISH_TOKEN=keyboardcat`                                              │ 416.2368ms    │ features/publish.feature:108                          │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────────────────────────────────────────────┤
│ /^I run cucumber-js with arguments `(|.+)` and env `(|.+)`$/                                                   │ 396.73ms      │ features/step_definitions/cli_steps.ts:29             │
│   I run cucumber-js with arguments `--publish` and env ``                                                      │ 613.364ms     │ features/publish.feature:121                          │
│   I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_ENABLED=true`                                  │ 492.013ms     │ features/publish.feature:121                          │
│   I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_TOKEN=f318d9ec-5a3d-4727-adec-bd7b69e2edd3`    │ 453.223999ms  │ features/publish.feature:58                           │
│   I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_ENABLED=1`                                     │ 446.359299ms  │ features/publish.feature:40                           │
│   I run cucumber-js with arguments `` and env `CUCUMBER_PUBLISH_TOKEN=f318d9ec-5a3d-4727-adec-bd7b69e2edd3`    │ 435.3693ms    │ features/publish.feature:121                          │
│   5 more                                                                                                       │               │                                                       │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────────────────────────────────────────────┤
│ /^I run cucumber-js \(installed (locally|globally)\)$/                                                         │ 253.48ms      │ features/step_definitions/cli_steps.ts:65             │
│   I run cucumber-js (installed locally)                                                                        │ 260.915699ms  │ features/global_install.feature:26                    │
│   I run cucumber-js (installed globally)                                                                       │ 246.0365ms    │ features/global_install.feature:17                    │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────┼───────────────┼───────────────────────────────────────────────────────┤
```

davidjgoss:
@aurelien-reeves thanks for posting - I think the durations look a little noisy there, it would benefit from maybe limiting to 2 or 3 decimal places in the formatter, but I don't think it's a huge deal.

davidjgoss (approved):

joac:
> @aurelien-reeves thanks for posting - I think the durations look a little noisy there, it would benefit from maybe limiting to 2 or 3 decimal places in the formatter, but I don't think it's a huge deal.

@davidjgoss Do you think that we should address that on this PR or a separate one?.

joac:
> As long as the measurement unit is the same (milliseconds I presume) I have no objections. Using a monotonic clock is the correct way to measure duration in any software system (using a wall clock is always wrong).

@aslakhellesoy  [From node the documentation](https://nodejs.org/api/perf_hooks.html#perf_hooks_performance_now): 
_Returns the current high resolution millisecond timestamp, where 0 represents the start of the current node process._
It didn't say anything about being monotonic, but [the implementation](https://github.com/nodejs/node/blob/master/lib/internal/perf/utils.js#L13) uses `process.hrtime()` [that claims](https://nodejs.org/api/process.html#process_process_hrtime_time) _These times are relative to an arbitrary time in the past, and not related to the time of day and therefore not subject to clock drift._ From that, I understand that is a monotonic clock.

davidjgoss:
@joac if you have time to do on this PR that’s be great!

joac:
@davidjgoss It's done!

davidjgoss (approved):

aslakhellesoy:
Hi @joac,

Aslak Hellesøy
Creator of Cucumber

PR #1801, opened 2021-09-27
debt: remove --predictable-ids option

davidjgoss:
# Description

- Remove the CLI option for `--predictable-ids` and related plumbing
- Make changes to fixture-based tests to ignore fields that will change on every run, as we do in the CCK tests (reusing the same list of keys)

# Motivation & context

We added predictable timestamps and identifiers at a time where we envisaged checking these values in the CCK so that all implementations would behave the same in this regard.

In practise, that level of parity was not practically achievable and so we now exclude timestamps and identifiers when validating against the CCK. Therefore it makes sense to remove this option and related complexity in the codebase.

## Type of change

(Technically a breaking change but there would be no sense in using this option except for our own testing. At any rate, it's going to be in a major version bump anyway.)

# Checklist:


aurelien-reeves (approved):
Approved

I would just wait for the review of @aslakhellesoy but looks ok to me 👌

aslakhellesoy (approved):
LGTM!

Issue #1845, opened 2021-11-19
Spaces in developer's path break Cucumber's own tests


mattwynne:
When working in the new contributors ensemble, we discovered that having a space in the path to where you're keeping cucumber-js's source code will break at least one of the tests.

Branch / draft-PR coming up!

## To reproduce

````
git clone git@github.com:cucumber:cucumber-js
mv cucumber-js "cucumber js"
cd "cucumber js"
npm install
npm test
````

## What happens

````
1) Scenario: Absolute path # features/formatter_paths.feature:26
   ✔ Before # features/support/hooks.ts:20
   ✔ Given a file named "features/a.feature" with: # features/step_definitions/file_steps.ts:10
       """
       Feature: some feature
         Scenario: some scenario
           Given a passing step
       """
   ✔ And a file named "features/step_definitions/cucumber_steps.js" with: # features/step_definitions/file_steps.ts:10
       """
       const {Given} = require('@cucumber/cucumber')

       Given(/^a passing step$/, function() {})
       """
   ✔ Given "{{{tmpDir}}}" is an absolute path # features/step_definitions/file_steps.ts:37
   ✔ When I run cucumber-js with `-f summary:{{{tmpDir}}}/summary.txt` # features/step_definitions/cli_steps.ts:19
   ✖ Then the file "{{{tmpDir}}}/summary.txt" has the text: # features/step_definitions/file_steps.ts:45
       """
       1 scenario (1 passed)
       1 step (1 passed)
       <duration-stat>
       """
       Error: ENOENT: no such file or directory, open '/Users/matt/git/github.com/cucumber/cucumber js/tmp/formatter_paths.feature_26/summary.txt'
   ✔ After # features/support/hooks.ts:104
````

mattwynne:
We did some digging on this today in the [new contributors ensemble](https://www.twitch.tv/videos/1209967860).

It turned out that the first thing we needed to do was wrap the path we pass to the formatter in the `When` step of the scenario in quotes, so that it wasn't split up into multiple tokens by [this call to `stringArgv`](https://github.com/cucumber/cucumber-js/blob/main/features/step_definitions/cli_steps.ts#L24).

However, once we'd done that, it looked like Cucumber was getting confused and no longer recognising it as an absolute path:

````
1) Scenario: Absolute path # features/formatter_paths.feature:26
   ✔ Before # features/support/hooks.ts:20
[...]

PR #1846, opened 2021-11-19
Handle spaces in dev paths (#1845)

eoola:
Issue #1845

added quotes to wrap the path to summary.txt
to ensure that paths containing spaces are read properly


mattwynne:
We botched the branching in this, so created #1847 instead.

Issue #1711, opened 2021-06-28
Proposal: Programmatic API for running cucumber-js


davidjgoss:
## Problem

Currently we don't have a good way to programmatically run cucumber-js. The need is from two angles:

- Testing - for the CCK and our acceptance tests in the project
- Projects that use cucumber-js as part of a framework (e.g. Serenity, Stryker)

## As-is

What tends to happen at the moment is a new instance of `Cli` is created with [strung-together argv input](https://github.com/cucumber/cucumber-js/blob/main/compatibility/cck_spec.ts#L41). It's obviously very unweildy and also [isn't on the public API]([https://github.com/cucumber/cucumber-js/issues/1710](https://github.com/cucumber/cucumber-js/issues/1710)).

Sometimes (possibly due to perceived fragility of the above) frameworks will just rely on the cucumber-js CLI but struggle to find ways to integrate and have their own options.

The [`Runtime` class](https://github.com/cucumber/cucumber-js/blob/main/src/runtime/index.ts#L42) is currently part of the public API but it's not useful in these contexts, depending on the pickles and support code to be provided by the caller.

## Proposal

Two components in the project:

### `runCucumber`

New async function that executes a test run in-process. Responsibilities:

- Accept an options object with a nice interface
- Do the "work" of resolving pickles, loading support code, running test cases, orchestrating formatters
- Return a promise that resolves to a result

This would be part of the public API and we'd encourage framework maintainers to use it when "wrapping" cucumber-js. We'd also use it for our own testing.

As much as possible, it would avoid direct interaction with `process`, instead accepting normalised options and stream interfaces for output, and leaving it to the caller to decide how to exit based on the result or an unhandled error.

Also `Runtime` should come off the public API as it's really an internal thing.

### CLI

Effectively a "client" of `runCucumber`. Responsibilities:

- Parse input from process arguments (via commander)
- Load profiles and mix those arguments in as needed
- Call `runCucumber` with the resolved options
- Exit as appropriate based on the results

This would continue to *not* be on the public API. Also it would only use functions/interfaces that *are* on the public API, such that we could easily break it out into its own package at some point, as is a common pattern now with projects like [Jest]([https://github.com/facebook/jest/tree/master/packages](https://github.com/facebook/jest/tree/master/packages)).

This decoupling also paves the way for some interesting new CLI features without having them bleed into the internals, e.g.:

- `--gui` for the cucumber-electron stuff
- `--interactive` for quick targeted reruns when TDD'ing

[...]

PR #1849, opened 2021-11-25
api: add runCucumber function internally

davidjgoss:
# Description

This adds the async `runCucumber` function for doing a test run programatically. It's not yet exposed publicly, but is used internally by the CCK tests.

This has involved:

- Pulling most of the guts out of the `Cli` class, since that is now a consumer of this new function
- Handling the i18n-related commands purely in the `cli` module
- Various other tweaks and refactors

# Motivation & context

First step towards #1711.

Now seems a good time to validate the approach on a few fronts:

- Defaulting of configuration options is generally moving _inwards_, so it's not tied to the argv parsing and so that API consumers can call with a very small signature if appropriate.
- The shape of the config object passed in to `runCucumber`. The idea is that the `support` block could be replaced with an existing support code library on subsequent runs in the same process, which solves a problem for a few framework authors.
- No breaking changes in the existing `Cli`, `Runtime` etc - framework authors are using these in the wild and we need to have a stable and compelling alternative before pulling them.

If we're happy and this goes in, the rest can follow quite quickly. Next steps:

- Implement the ability to reuse support code library
- Add a mechanism to load configuration from argv, files etc in a sane way
- Add integration tests specifically for this API
- Expose publicly (probably with an "experimental" qualifier), add documentation

## Type of change

# Checklist:


aurelien-reeves (commented):
After a first look, it looks really great! Well done!

I'll have a deeper look later

nicojs:
Greet to see progress here! 😍

> The shape of the config object passed in to `runCucumber`. The idea is that the `support` block could be replaced with an existing support code library on subsequent runs in the same process, which solves a problem for a few framework authors.

Good to hear it is still on the radar. It will dramatically improve mutation testing performance ⏩💨

src/run/runCucumber.ts
```diff
@@ -0,0 +1,103 @@
...
+import { IRunConfiguration } from '../configuration'
+import { IRunResult } from './types'
+import { resolvePaths } from './paths'
+import { makeRuntime } from './runtime'
+import { initializeFormatters } from './formatters'
+import { getSupportCodeLibrary } from './support'
+
+export async function runCucumber(
+  configuration: IRunConfiguration,
+  environment: {
```
mattwynne, line 21: Can we name this type?

mattwynne:
I can only offer nit-picks, this looks great to me!

aurelien-reeves (approved):
It looks really great!
I agree with Matt's comment.
Beside that, this is an approval 🚀

davidjgoss:
Thanks for the feedback @aurelien-reeves @nicojs @mattwynne - I'm going to address Matt's very fair nit-pick, beef up the tests a bit and then get this merged and on to the next part.

nicojs:
Q: does this break the current (undocumented) API? 

This is what we're using in our cucumber plugin right now:

```ts
const argv = [
      'node',
      'cucumber-js',
      '--retry',
      '0',
      '--parallel',
      '0',
      '--format',
      require.resolve('./stryker-formatter'),
      ...bailArgs,
      ...tagsArgs,
      ...profileArgs,
      ...testFilterArgs,
    ];
    const cli = new Cli({
      argv,
      cwd: process.cwd(),
      stdout: process.stdout,
    });
   await cli.run();
```

aurelien-reeves:
> Question: does this break the current (undocumented) API?
> 
> This is what we're using in our cucumber plugin right now:
> 
> ```ts
> const argv = [
>       'node',
>       'cucumber-js',
>       '--retry',
>       '0',
>       '--parallel',
>       '0',
>       '--format',
>       require.resolve('./stryker-formatter'),
>       ...bailArgs,
>       ...tagsArgs,
>       ...profileArgs,
>       ...testFilterArgs,
>     ];
>     const cli = new Cli({
>       argv,
>       cwd: process.cwd(),
>       stdout: process.stdout,
>     });
>    await cli.run();
> ```

I took a look at the code in the branch. As far as I can tell, no, your snippet won't be broken. It will use the new API under the hood, but it will work as expected :)

davidjgoss:
@nicojs @aurelien-reeves yep, very much a conscious choice to avoid any breaking changes to existing Cli and Runtime for the near to mid term at least.

davidjgoss:
@binarymist I don't think `getConfiguration` was intended to be public, it perhaps should have been a private method. I think I understand your use case though having looked through the code and this should be catered for via the new API in the next RC.

binarymist:
> this should be catered for via the new API in the next RC

Great!

Problme is... If we didn't use `getConfiguration`, then we wouldn't be able to show the _Build User_ (think end user) the test plan that they are about to run.

PR #1862, opened 2021-12-01
Update esm.md

aurelien-reeves:
# Description

Move the chapter about configuration file at the top of the doc to give it more visibility

# Motivation & context

The fact that the cucumber configuration file must remain a commonJs file is very important as without that, we may think the esm support is actually broken.

Moving the chapter back on top should give it more visibility.


davidjgoss (approved):
Good catch, thanks.

Issue #1852, opened 2021-11-26
Feature Tests Failing on Windows


gloryCU:
**Describe the bug**
While pairing with @mattwynne to set up my development environment, we found that the scenarios all failed. Same error: 
```
293) Scenario: custom world constructor is passed the parameters # features\world_parameters.feature:76
     × Before # features\support\hooks.ts:20
         Error: EPERM: operation not permitted, symlink 'C:\Users\gugwu\Documents\OSContribution\cucumber-js' -> 'C:\Users\gugwu\Documents\OSContribution\cucumber-js\tmp\world_parameters.feature_76\node_modules\@cucumber\cucumber'
             at Object.symlinkSync (fs.js:1210:3)
             at Object.createSymlinkSync (C:\Users\gugwu\Documents\OSContribution\cucumber-js\node_modules\fs-extra\lib\ensure\symlink.js:76:13)             at World.<anonymous> (C:\Users\gugwu\Documents\OSContribution\cucumber-js\features\support\hooks.ts:42:11)
             at Object.run (C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\user_code_runner.ts:37:21)
             at Object.run (C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\runtime\step_runner.ts:50:41)
             at processTicksAndRejections (internal/process/task_queues.js:95:5)
             at async TestCaseRunner.invokeStep (C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\runtime\test_case_runner.ts:130:12)    
             at async TestCaseRunner.runHook (C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\runtime\test_case_runner.ts:246:12)       
             at async C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\runtime\test_case_runner.ts:201:20
             at async TestCaseRunner.aroundTestStep (C:\Users\gugwu\Documents\OSContribution\cucumber-js\src\runtime\test_case_runner.ts:162:28)     - Given a file named "features/passing_steps.feature" with: # features\step_definitions\file_steps.ts:10
         """
         Feature: a feature
           Scenario: a scenario
             Given the world parameters are correct
         """
     - Given a file named "features/support/world.js" with: # features\step_definitions\file_steps.ts:10
         """
         const {setWorldConstructor} = require('@cucumber/cucumber')    

         function CustomWorld(options) {
           for(const key in options.parameters) {
             this[key] = options.parameters[key]
[...]

PR #1863, opened 2021-12-01
Smoother onboarding for Windows developers

mattwynne:
# Description

* Add a try/catch block around a line in our `features/support` code that creates a symlink. If you're using Windows, unless you have enabled "Developer Mode" this line throws an error. We now catch the error and show a helpful message to the user.
* Replace uses of `cp` command (which is not available in a Windows PowerShell / `cmd.exe` prompt) with an equivalent `npx` command that will work cross-platform. 

# Motivation & context

* The features won't run if you don't have "Developer Mode" enabled. See #1852
* The `build-local` script won't run unless you have a `cp` command in your path.

Big thanks to @RukenaG for inspiring this change!

## Type of change

# Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -257,6 +257,7 @@
     "prettier": "2.5.0",
     "reindent-template-literals": "1.1.0",
     "semver": "7.3.5",
+    "shx": "^0.3.3",
     "sinon": "12.0.1",
     "sinon-chai": "3.7.0",
     "stream-buffers": "3.0.2",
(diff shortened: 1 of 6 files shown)


mattwynne:
@aurelien-reeves I tried `cp-cli` first but it didn't seem to work; `shx` seems a bit more mature, and works!

davidjgoss (approved):

package.json
```diff
@@ -266,7 +267,7 @@
     "typescript": "4.5.2"
   },
   "scripts": {
-    "build-local": "tsc --build tsconfig.node.json && cp src/importer.js lib/ && cp src/wrapper.mjs lib/",
+    "build-local": "tsc --build tsconfig.node.json && npx shx cp src/importer.js lib/ && npx shx cp src/wrapper.mjs lib/",
```
aurelien-reeves, line 270: I think the `npx` prefix is not required as part of npm scripts

aurelien-reeves (approved):

Issue #1853, opened 2021-11-30
Expose Cucumber version


jan-molak:
**Is your feature request related to a problem? Please describe.**

[Serenity/JS](https://serenity-js.org) currently supports Cucumber versions 1.x to 7.x and I'd like it to be able to support version 8.x as well.

In order to do that, Serenity/JS needs to know what version of Cucumber it's being used with.

At the moment, Serenity/JS figures out the version based on the `version` field present in Cucumber's `package.json` file.

However, in Cucumber 8 RC1 `package.json` is not listed in `exports`, so trying to `require('@cucumber/cucumber/package.json')` results in:

```
 Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports"
```

**Describe the solution you'd like**

It would be great if Cucumber 8 could either:
- added `package.json` to its exports so that Serenity/JS could read the version from it
- exported a `version` const so that it could be imported by consumer frameworks (e.g. `import { version } from '@cucumber/cucumber'`)

**Describe alternatives you've considered**

An alternative solution would be for Serenity/JS to check if `@cucumber/cucumber` module is present, and if it is and it's not possible to load its `package.json` => assume it's version 8.

It _could_ work, but it smells of a an awful hack ;-)

**Additional context**
N/A

aurelien-reeves:
Indeed, we should try to find a way before releasing final version of 8.0.0

Thanks for the report!

jan-molak:
Thanks, give me a shout if you'd like me to try it again with any new RCs.

jan-molak:
Nice one, thanks @aurelien-reeves, #1866 looks great.
I'm invoking Cucumber.js [programmatically](https://github.com/serenity-js/serenity-js/blob/0c706f753343f1ce35b7e77460062d5762f6d4aa/packages/cucumber/src/cli/CucumberCLIAdapter.ts#L101-L117) rather than spawning a new process, so importing a `version` is preferable.

PR #1866, opened 2021-12-02
Export version number of cucumber-js

aurelien-reeves:
# Description

Fixes. #1853 
With ESM support, the version number of cucumber-js is not as exposed as it was due to package.json not exported.
It seems more reliable with ESM to have version defined in a source file then exported by the libraries.

To keep the version sync with the one defined in package.json, the npm package [genversion](https://www.npmjs.com/package/genversion) has been used.

# Motivation & context

Pretty sure it may be useful with the incoming API too :)

## Type of change

I think we could consider this a breaking change: the way to find the version of cucumber programmatically has changed

- Breaking change (will cause existing functionality to not
  work as expected)

# Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ usage.txt
 yarn-error.log
 .vscode
 .DS_Store
+src/version.ts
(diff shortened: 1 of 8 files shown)


davidjgoss (approved):
Neat.

aurelien-reeves:
Do we need to add some tests for this?

PR #1870, opened 2021-12-10
Add package.json to node module exports

### LINKED ISSUE #1869: [ERR_PACKAGE_PATH_NOT_EXPORTED] when running ng e2e
Hello, trying to use cucumber with protractor in angular.
When running `ng e2e --port 4201` I get an error:

E/launcher - Error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED] [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './package.json' is not defined by "exports" in C:\Users\dndum\Documents\Projects\angular-project\node_modules\@cucumber\cucumber\package.json
    at throwExportsNotFound (internal/modules/esm/resolve.js:299:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:522:3)
    at resolveExports (internal/modules/cjs/loader.js:424:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:464:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:802:27)
    at Function.Module._load (internal/modules/cjs/loader.js:667:27)
    at Module.require (internal/modules/cjs/loader.js:887:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at ModuleLoader.require (C:\Users\dndum\Documents\Projects\angular-project\node_modules\@serenity-js\core\src\io\ModuleLoader.ts:68:20)
    at ModuleLoader.versionOf (C:\Users\dndum\Documents\Projects\angular-project\node_modules\@serenity-js\core\src\io\ModuleLoader.ts:80:33)
[11:27:45] E/launcher - Process exited with error code 100
An unexpected error occurred: undefined

Please advise on how to proceed.
Thank you!

aurelien-reeves:
# Description

Attempt to fix #1869

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [Unreleased]
 ### Fixed
+- Add `package.json` to node module `exports`
+  ([PR#1870](https://github.com/cucumber/cucumber-js/pull/1870)
+   [Issue#1869](https://github.com/cucumber/cucumber-js/issues/1869))
 - Allows for parentheses in paths for developers working on cucumber's own code ([[#1735](https://github.com/cucumber/cucumber-js/issues/1735)])
 - Smoother onboarding for Windows developers ([#1863](https://github.com/cucumber/cucumber-js/pull/1863))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [Unreleased]
 ### Fixed
-- Add `package.json` to node module `exports`
+- Ensure package.json can be imported by consuming projects
   ([PR#1870](https://github.com/cucumber/cucumber-js/pull/1870)
    [Issue#1869](https://github.com/cucumber/cucumber-js/issues/1869))
 - Allows for parentheses in paths for developers working on cucumber's own code ([[#1735](https://github.com/cucumber/cucumber-js/issues/1735)])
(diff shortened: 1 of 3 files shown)


davidjgoss (approved):

aurelien-reeves:
Do we keep that one: https://github.com/cucumber/cucumber-js/pull/1866?

I would keep it. I think exporting the version number can be useful. But this remain duplication so, what do you think?

davidjgoss (commented):
> I would keep it. I think exporting the version number can be useful. But this remain duplication so, what do you think?

No problem with having both from my side (this is the more important because of existing usage, but I think having an export for it makes sense too).

Also, could we add a test to prove this works? There is one here that covers the paths under `lib` that we could extend:

https://github.com/cucumber/cucumber-js/blob/main/features/direct_imports.feature#L48

aurelien-reeves:
> > I would keep it. I think exporting the version number can be useful. But this remain duplication so, what do you think?
> 
> No problem with having both from my side (this is the important one because of existing usage, but I think having an export for it makes sense too).
> 
> Also, could we add a test to prove this works? There is one here that covers the paths under `lib` that we could extend:
> 
> https://github.com/cucumber/cucumber-js/blob/main/features/direct_imports.feature#L48

I will merge the PR wich export the version as part of the API, then add a test in direct_imports.feature which will then cover both:
- the possibility to import the version number from the package.json
- the possibility to import the version number from the library

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [Unreleased]
 ### Fixed
+- Add `package.json` to node module `exports`
```
davidjgoss, line 12: To be clearer I'd maybe say "Ensure `package.json` can be imported by consuming projects"

davidjgoss (approved):

PR #1889, opened 2022-01-11
Add a simple TypeScript example project

mattwynne:
# Description

Add a simple example project for people using TypeScript

# Motivation & context

We've heard from our users that they'd like more examples, and better documentation.

Today, someone asked on Slack about how to test a TypeScript project. Rather than explain it to them, I thought I'd show them an example.

I'm hoping we can add more of these, for other typical use-cases.

## Type of change

# Questions / puzzles / feedback wanted

* Should we run this example as part of the CI build, to check for regressions?
* Should we reference the version of Cucumber in the project itself (i.e. use a relative `file://..` path in `package.json` or continue using a dependency on a stable / released version of Cucumber?
* Is there anything about this example that could be more exemplary, or could be removed to make it simpler?
* I've added the folder to the list of exclusions in `dependency-lint.yml` because it was giving an error - was that the right thing to do?


davidjgoss (commented):
Thanks @mattwynne, I think we've been sorely lacking some solid examples for a while now.

I'm not totally sure but I'm tending towards wanting these examples in a separate repo so we don't have to uncouple them from the tooling in this project (linting etc as you've found). That repo could also serve as a list of community-sourced examples and templates e.g. https://github.com/davidjgoss/cucumber-puppeteer-typescript-starter (sorry for the plug).

Also I'm hoping this is a step towards something like:

```
npx cucumber-js init --typescript
```

(By the way, I'm still working at the documetation - more good stuff coming soon on that front including a better README.)

examples/typescript/features/greeting.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Greeing
```
davidjgoss, line 1: typo -> Greeting

mattwynne:
> I'm not totally sure but I'm tending towards wanting these examples in a separate repo so we don't have to uncouple them from the tooling in this project (linting etc as you've found). That repo could also serve as a list of community-sourced examples and templates e.g. https://github.com/davidjgoss/cucumber-puppeteer-typescript-starter (sorry for the plug).

Yep, that would work too. I guess I'm thinking about discover-ability by placing it here, but we could still have an `examples` folder in this repo that just contains a README linking out to other places.

Were you thinking one repo per example or one repo _of_ examples?

davidjgoss:
> Were you thinking one repo per example or one repo of examples?

Good question. I had been thinking one repo, but if we did them individually then they could be [template repos](Were you thinking one repo per example or one repo of examples?) - not sure how commonly these would be used though.

> I guess I'm thinking about discover-ability by placing it here

On this point I think the list of examples should go in the readme, probably right below the `npm install` instructions, like "clone one of these examples and read on for the full docs" kind of thing. What do you think?

aslakhellesoy:
Should this go into https://github.com/cucumber-examples/ ?

davidjgoss:
I think that would work, with a link from the readme in this repo. Matt?

mattwynne:
> Should this go into https://github.com/cucumber-examples/ ?

I didn't know about that org!

Honestly, that feels a bit far away to me. I'd prefer them to be at least under the cucumber org for discover-ability.

mattwynne:
@aslakhellesoy do you know how come those example repos are not under the https://github.com/cucumber-school org? Aren't they used for in-person school trainings?

mattwynne:
If we use one repo for all examples, I think that will make it more likely that we'll get community submissions (as they can be made via pull requests to that repo). I think that probably trumps the advantage of the template repo option, WDYT?

mattwynne:
OK, I've moved it into https://github.com/cucumber-examples/cucumber-js-examples

We should probably think about how to have some kind of CI on that repo so that it doesn't rot.

mattwynne:
Questions/actions before we close this PR:

* Should we move that repo under the cucumber org or is it fine where it is?
* Where are the places we should link to that examples repo so it's easily discover-able?

jenisys:
* I think "cucumber-examples" is a good place to look for examples how to use cucumber.
* One place in the docs to point to it could be an item/link on the https://cucumber.io/docs/guides/ page (because it is similar/related to the "10 Minute Tutorial").

mattwynne:
Ok, it's now live here: https://github.com/cucumber-examples/cucumber-js-examples

mattwynne:
I'm still wondering about where's the best place to put signposts so these are easy to discover.

I noticed we have https://github.com/cucumber/cucumber-js/blob/main/docs/nodejs_example.md linked from the README. Maybe we should change the README to just link to the examples repo and remove this page?

aslakhellesoy:
> @aslakhellesoy do you know how come those example repos are not under the https://github.com/cucumber-school org? Aren't they used for in-person school trainings?

IIRC it was @sebrose who created the `cucumber-examples` org. Seb, can you refresh our memory about why we didn’t just add more repos to the `cucumber-school` org instead?

aurelien-reeves:
> I'm still wondering about where's the best place to put signposts so these are easy to discover.
> 
> I noticed we have https://github.com/cucumber/cucumber-js/blob/main/docs/nodejs_example.md linked from the README. Maybe we should change the README to just link to the examples repo and remove this page?

👍
A list of examples in the readme would be nice IMO

davidjgoss:
Starting point: https://github.com/cucumber/cucumber-js/pull/1918 - we could list the examples there or just link to the examples root, I don't feel strongly either way.

PR #1893, opened 2022-01-14
Extract prettier config from eslintrc

mattwynne:
# Description

Use a more conventional place to store prettier config

# Motivation & context

So that VSCode's prettier plugin can automatically find it.

Currently, VSCode doesn't understand our prettier config, so the code formatter from the prettier plugin reformats the code using defaults (e.g. adding semi-colons), which then breaks the eslint rules!

## Type of change

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -1,26 +1,24 @@
 extends:
-  - "eslint:recommended"
-  - "plugin:import/typescript"
-  - "plugin:@typescript-eslint/eslint-recommended"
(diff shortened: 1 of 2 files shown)


mattwynne:
@davidjgoss can you give this a quick spin on WebStorm to make sure it all works OK for you? I guess you may need to change to using prettier for linting? As far as I can tell, the `lint-autofix` npm task still works, so I _think_ eslint is still finding the config.

davidjgoss (approved):
LGTM!

PR #1905, opened 2022-01-31
build: tag as next when publishing release candidates

davidjgoss:
# Description

Creates two versions of the npm release workflow - one for stable releases, and one for release candidates, based on branch naming patterns. The former are tagged as `latest` and the latter as `next`.

# Motivation & context

Unless a tag is specified, npm packages default to `latest` on publication. This meant that, when a release candidate was out, in practise users would run `npm install @cucumber/cucumber` and get the release candidate when we would want them to get the latest stable version unless they opt in to the RC.

`next` is a loose convention in the npm ecosystem for tagging prereleases. So next time there is an RC, regular users will still get the latest stable version when they install, and enthusiasts can do `npm install @cucumber/cucumber@next` to get the RC.

## Type of change

# Checklist:

- [ ] My change needed additional tests

### DIFF
--- a/.github/workflows/release-npm-latest.yaml
+++ b/.github/workflows/release-npm-latest.yaml
@@ -1,0 +1,24 @@
+name: Release NPM (latest)
+
+on:
+  push:
+    branches: [release/*]
+    branches-ignore: [release/*-rc*]
+
+jobs:
+  publish-npm:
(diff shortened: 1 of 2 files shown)


mattwynne:
Lovely!

PR #1922, opened 2022-02-08
Raise an error when the version of node is older than package.json's engines.node

### LINKED ISSUE #1890: Warn users on old (pre-ESM) node versions
**Describe the bug**
A user on node v11.15.0 got the error "Not Supported" from line 10 of `src/importer.js`

**To Reproduce**
1. Install / switch to node v11.15.0
2. Make a simple project with a step definition
3. Run it

**Expected behavior**
I think we should check the `process.version` and let the user know that they'll need to upgrade node to use this version of Cucumber. The "Not supported" error coming from within the bowels of Cucumber is not helpful or actionable neough.

**Additional context**
Screenshot 2022-01-12 at 8 28 25 AM

mattwynne:
# Description

Adds a warning if the `process.version` is too old for Cucumber to work.

# Motivation & context

Fixes #1890 - we had a user who was running an older node version and got confused.

## Type of change

- Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [x] My change needed additional tests


mattwynne:
It's nigh-on impossible to test this other than manually. Here's the result of my manual test:

```
➜  cucumber-js git:(check-node-version) node --version
v11.10.1
➜  cucumber-js git:(check-node-version) npm run feature-test
(node:57566) ExperimentalWarning: The fs.promises API is experimental

> @cucumber/cucumber@8.0.0-rc.2 prefeature-test
> npm run build-local


> @cucumber/cucumber@8.0.0-rc.2 build-local /Users/matt/git/github.com/cucumber/cucumber-js
> genversion --es6 src/version.ts && tsc --build tsconfig.node.json && shx cp src/importer.js lib/ && shx cp src/wrapper.mjs lib/


> @cucumber/cucumber@8.0.0-rc.2 feature-test
> node ./bin/cucumber-js

internal/modules/cjs/loader.js:615
    throw err;
    ^

Error: Cannot find module '/Users/matt/git/github.com/cucumber/cucumber-js/lib/check_node_version.js'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:613:15)
    at Function.Module._load (internal/modules/cjs/loader.js:539:25)
    at Module.require (internal/modules/cjs/loader.js:667:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (/Users/matt/git/github.com/cucumber/cucumber-js/lib/importer.js:1:63)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
```

mattwynne:
Two possible improvements:

1. It might be nicer to print the error message without the stacktrace, but I'm not sure how to do that without adding console-specific code, which feels wrong in this file. (I'm kinda surprised there's not an error handler at the UI level?)

2. It seems like putting the check in the `importer` is the right place to have it, since that's where we need to be on that version, but I guess it could move more centrally, and look up the right version from the `package.json`?

davidjgoss (changes requested):
> It might be nicer to print the error message without the stacktrace, but I'm not sure how to do that without adding console-specific code, which feels wrong in this file.
> 
> It seems like putting the check in the importer is the right place to have it, since that's where we need to be on that version, but I guess it could move more centrally

Lack of ESM support is the symptom you ran into, but I think there's a good chance of us using some other language feature or standard library that's only in our supported versions. I'd advocate for putting it here https://github.com/cucumber/cucumber-js/blob/main/src/cli/run.ts which is the entry point for the CLI, and where you could just do `console.error` and then `process.exit(1)` because as you've noted the stack trace isn't useful in this scenario.

> ...and look up the right version from the package.json?

Nice to have, but not essential IMO. I think it would be sufficient to add a section to the `CONTRIBUTING.md` on how to manage dropping an EOL'd Node version.

src/importer.js
```diff
@@ -11,3 +11,11 @@ async function importer(descriptor) {
 }
 
 module.exports = { importer }
+
+// None of this stuff will work on versions of Node older than v12
+const MIN_NODE_VERSION = 'v12'
+if (process.version < MIN_NODE_VERSION) {
+  throw new Error(
+    `Cucumber can't run on this version (${process.version}) of Node. Please upgrade to at least ${MIN_NODE_VERSION}.`
```
aslakhellesoy, line 19: Suggested edit:

`Cucumber can only run on Node.js versions ${MIN_NODE_VERSION} and greater. (This Node.js version is ${process.version}).`

(It's called Node.js, not Node).

aslakhellesoy (changes requested):

src/importer.js
mattwynne: I was trying to find a way to craft the sentence so you didn't have a full-stop after "Node.js" which looked weird. Thanks!

mattwynne:
OK this is good enough now IMO. Please check and merge!

src/cli/assert_node_engine_version.ts
```diff
@@ -0,0 +1,25 @@
...
+const readActualPackageJSON: () => PackageJSON = () =>
+  JSON.parse(
+    fs
+      .readFileSync(path.resolve(__dirname, '..', '..', 'package.json'))
+      .toString()
+  )
+
+export function assertNodeEngineVersion(
+  currentVersion: string,
+  readPackageJSON: () => PackageJSON = readActualPackageJSON
```
aurelien-reeves, line 18: Is this second parameter for mocking purpose?
Do we actually need this?

davidjgoss (approved):

src/cli/assert_node_engine_version.ts
aurelien-reeves: Yes, this is used
but is it useful?

mattwynne: yep, it's just to de-couple it. It felt wrong to have the function be responsible for reading from the file.

PR #1924, opened 2022-02-16
avoid throwing for flow control

charlierudolph:
Instead of having a try/catch around this function just have it return an optional error message.

### DIFF
--- a/src/cli/assert_node_engine_version_spec.ts
+++ b/src/cli/assert_node_engine_version_spec.ts
@@ -14,7 +14,9 @@ describe(validateNodeEngineVersion.name, () => {
     }))
 
     // Assert
-    expect(error).to.eql("Cucumber can only run on Node.js versions >=12. This Node.js version is 11.0.0")
+    expect(error).to.eql(
+      'Cucumber can only run on Node.js versions >=12. This Node.js version is 11.0.0'
+    )
   })
 
   it('returns null when the version is greater than specified in package.json', () => {
--- a/src/cli/assert_node_engine_version_spec.ts
+++ b/src/cli/assert_node_engine_version_spec.ts
@@ -1,35 +1,37 @@
-import assert from 'assert'
 import { expect } from 'chai'
+import * as sinon from 'sinon'
 import { validateNodeEngineVersion } from './assert_node_engine_version'
 
 describe(validateNodeEngineVersion.name, () => {
   it('returns an error message when the version is lower than specified in package.json', () => {
     // Arrange
+    const errorSpy = sinon.spy()
 
     // Act
-    const error = validateNodeEngineVersion('v11.0.0', () => ({
+    validateNodeEngineVersion('v11.0.0', errorSpy, () => ({
       engines: {
         node: '>=12',
       },
     }))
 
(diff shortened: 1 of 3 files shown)


src/cli/assert_node_engine_version_spec.ts
```diff
@@ -1,18 +1,33 @@
...
-    assertNodeEngineVersion('v17.0.0')
+  it('returns null when the version is greater than specified in package.json', () => {
+    // Arrange
+
+    // Act
+    const result = validateNodeEngineVersion('v17.0.0', () => ({
+      engines: {
+        node: '>=12',
+      },
+    }))
```
charlierudolph, line 28: Think its better to hardcode the engines here as well so this doesn't suddenly fail when we no longer support version 17 (probably a little ways off)

aurelien-reeves (changes requested):
Avoiding throwing an error can be a good idea 👍

But rather than returning `string | null` - which actually I would have expect to be the return type of the validateNodeEngineVersion  function- with the string being kinda error message, would it be possible to return something that would look like a result struct which could looks like the following?

```typescript
{
  status: bool,
  message: string?
}
```

Or, just, a boolean, and the caller could be responsible for the error message?

src/cli/assert_node_engine_version_spec.ts
mattwynne: 💯, I missed that.

mattwynne:
I did consider using a promise as a kind of Maybe object, so the calling code could do something like

```
validateNodeEngineVersion(process.version).catch(errorMessage => {
  console.error(error) // eslint-disable-line no-console
  process.exit(1)
})
```

Still using exceptions for flow control really, but perhaps it reads better?

Another alternative would be to pass the assertion function a callback to run if the engine is invalid, like:

```
validateNodeEngineVersion(process.version, errorMessage => {
  console.error(error) // eslint-disable-line no-console
  process.exit(1)
})
```

That way the conditional stays inside the function. That's probably my favourite of them all.

src/cli/assert_node_engine_version_spec.ts
davidjgoss: Oh true, the list right now is 12, 14, 16, 17

davidjgoss:
@mattwynne I quite like the callback idea there.

davidjgoss (approved):

mattwynne (approved):
lovely!

src/cli/assert_node_engine_version_spec.ts
```diff
@@ -1,18 +1,37 @@
...
+
+    // Assert
+    expect(errorSpy).to.have.been.calledOnceWith(
+      'Cucumber can only run on Node.js versions >=12. This Node.js version is v11.0.0'
     )
   })
 
-  it('passes when the version is greater than specified in package.json', () => {
-    assertNodeEngineVersion('v17.0.0')
+  it('returns null when the version is greater than specified in package.json', () => {
```
mattwynne, line 23: I just realised we should have changed these example descriptions too: we're not returning anything anymore.

Issue #1871, opened 2021-12-10
Change Cucumber's own features to use cucumber-expressions (from regexes)


mattwynne:
Right now we're using regular expressions in our own step definitions.

It would be good to change these over to using cucumber expressions (and parameter types!) so that we're providing a good example of how Cucumber should be used.

davidjgoss:
Agreed!

aslakhellesoy:
If you're mobbing with new contributors on this - show them https://cucumber.github.io/cucumber-expressions/ !

shvmsnju:
Hi @mattwynne, I would like to contribute to this issue.

mattwynne:
Awesome @shvmsnju, let me know if you need any help.

pk:
Is this issues still valid/pending? I could have a look to get started with the code base.

davidjgoss:
@pk absolutely, feel free to raise PRs and let me know if you need any pointers.

pk:
Perfect, we are using Cucumber so much that we really would like to contribute. Also our test suites are getting more and more complex and we may need to start to use some features from API, running the Features dynamically so it would be nice to contribute before I start asking too many questions. ;)

pk:
OK, I've looked into this and I think this task is done.

There are steps under `compability/**/*.ts` and `features/step_definitions/*.ts` all of which has been converted into the Cucumber Expressions.

I've done search for Given/When/Then with the RegExp patterns and it really yields references to the steps with RegExp only in *.md files.

I may be missing something....

PR #1926, opened 2022-02-19
Regexp to cucumberexp

shvmsnju:
### 🤔 What's changed?

Changed 3 instances of regular expressions in the step definitions with cucumber expressions

The change is a small step in fixing issue #1871 

### 📋 Checklist:


shvmsnju:
Hi @mattwynne, in the cli.steps.ts file in step-definitions, 
when i try to change  When(  /^I run cucumber-js \(installed (locally|globally)\)$/ ........
to When('I run cucumber-js \(installed (locally|globally))' it is showing linting problem (no-useless-escapes) 
How do we change this kind of regexp to cucumber expressions

davidjgoss:
> If you look at [the step definition](https://github.com/cucumber/cucumber-js/blob/main/features/step_definitions/cli_steps.ts#L83) it just has a big if statement around it anyway, so I'd just split this into two step definitions each with a hard-coded locally or globally.

+1 I think this is the cleanest way.

mattwynne:
@shvmsnju since this stands as a good change on its own, I'll merge it in and you can deal with [this issue](https://github.com/cucumber/cucumber-js/pull/1926#issuecomment-1045963883) in a separate PR, OK?

mattwynne (approved):

shvmsnju:
> @shvmsnju since this stands as a good change on its own, I'll merge it in and you can deal with [this issue](https://github.com/cucumber/cucumber-js/pull/1926#issuecomment-1045963883) in a separate PR, OK?

@mattwynne yes of course! I was just a busy with work so i was not able to do that untill now. Will try to do that during this weekend!

shvmsnju:
Hi @mattwynne,

Even if we break the step definition into two parts, When we implement the step definition like: 

When('I run cucumber-js \(installed locally\)',{ 

Eslint will not allow escaping the bracket using \ as we don't need it in regular string and eslint considers this as a useless escape character and removes it when we save the file. To escape the bracket so that it is considered as a literal bracket and not optional text in cucumber expression we need to escape the \ character before the bracket using another \. I guess the whole thing sounds pretty confusing but you will understand when u see the code. I ran the tests and they are passing.

PR #1931, opened 2022-02-23
esm: use specific option for import vs require for support code

davidjgoss:
### 🤔 Motivation and changes

As part of the initial work to support ESM, I made it so that all support code was `import`ed by default, except sometimes --- based on heuristics around file extension and whether `--require-module` had been specified --- it fell back to `require`. It was a kind of aggressive move towards the new module system.

On reflection I think this was a mistake. The heuristics are bound to be wrong some of the time and potentially leave users with a breaking change that's difficult to work around. A future major release where we go ESM-only is probably not far away and so it makes sense to drop `require` entirely at that point.

Fortunately the ESM support has only gone as far as the 8.x RC so it's not too late to pivot. This PR makes it so that:

- `--require` does the same as it always did before, no breaking changes
- `--import`, a new option, works like `--require` but imports instead, so we'd expect ESM projects taking advantage of the new support in 8.x to use this
- If neither `--require` or `--import` are specified, the automatic `require`ing does what it did before based on feature paths, but also any `.mjs` files found in the same way are `import`ed - this is a backwards-compatible change because `.mjs` files didn't work at all previously

_Not_ rowing back on formatters and snippets using `import` - I think this still right because it's much more predictable and there's no just-in-time transpilation involved to complicate things.

### 📋 Checklist:


davidjgoss:
Still need to add and update some docs but feedback on the what and why would be great.

src/run/paths.ts
```diff
@@ -114,3 +113,32 @@ async function expandFeaturePaths(
...
+  requirePaths: string[]
+  importPaths: string[]
+}> {
+  if (
+    unexpandedRequirePaths.length === 0 &&
+    unexpandedImportPaths.length === 0
+  ) {
+    const defaultPaths = getFeatureDirectoryPaths(cwd, featurePaths)
+    const requirePaths = await expandPaths(cwd, defaultPaths, '.js')
+    const importPaths = await expandPaths(cwd, defaultPaths, '.mjs')
```
aurelien-reeves, line 132: What about also supporting `.cjs` extension?

aurelien-reeves (commented):
Looks good

src/run/paths.ts
davidjgoss: Oh good point, that's still backwards-compatible as a default. Done!

davidjgoss:
No objections so far; I'll update the documentation tonight and merge tomorrow.

docs/cli.md
```diff
@@ -26,19 +26,24 @@ needs to be required in your support files and globally installed modules cannot
...
+By default, the following files are loaded:
 * If the features live in a `features` directory (at any level)
-  * `features/**/*.js`
+  * `features/**/*.(js|cjs|mjs)`
 * Otherwise
-  * `<DIR>/**/*.js` for each directory containing the selected features
+  * `<DIR>/**/*.(js|cjs|mjs)` for each directory containing the selected features
 
-Alternatively, you can use `--require <GLOB|DIR|FILE>` to explicitly require support files before executing the features. Uses [glob](https://github.com/isaacs/node-glob) patterns.
+With the defaults described above, `.js` files are loaded via `require()`, whereas `.cjs` and `.mjs` files are loaded via `import()`.
```
aurelien-reeves, line 37: `.cjs` files shouldn't be loaded using `require()` too?

docs/cli.md
```diff
@@ -26,19 +26,24 @@ needs to be required in your support files and globally installed modules cannot
...
-This option may be used multiple times in order to e.g. require files from several different locations.
+Alternatively, you can use either or both of these options to explicitly load support files before executing the features:
 
-_Note that once you specify any `--require` options, the defaults described above are no longer applied._
+- `--require <GLOB|DIR|FILE>` - loads via `require()` (legacy)
+- `--import <GLOB|DIR|FILE>` - loads via `import()`
+
+Both options use [glob](https://github.com/isaacs/node-glob) patterns and may be used multiple times in order to e.g. load files from several different locations.
+
+_Note that once you specify any `--require` or `--import` options, the defaults described above are no longer applied._
```
aurelien-reeves, line 46: This looks important. I'll move it at the top of the "Loading support files" chapter

docs/esm.md
```diff
@@ -1,6 +1,8 @@
 # ES Modules (experimental)
```
aurelien-reeves, line 1: Do we still consider esm support experimental?

src/run/paths.ts
```diff
@@ -114,3 +113,32 @@ async function expandFeaturePaths(
...
+  requirePaths: string[]
+  importPaths: string[]
+}> {
+  if (
+    unexpandedRequirePaths.length === 0 &&
+    unexpandedImportPaths.length === 0
+  ) {
+    const defaultPaths = getFeatureDirectoryPaths(cwd, featurePaths)
+    const requirePaths = await expandPaths(cwd, defaultPaths, '.js')
+    const importPaths = await expandPaths(cwd, defaultPaths, '.@(cjs|mjs)')
```
aurelien-reeves, line 132: I would have consider `.cjs` extension as a CommonJs file loaded using `require` rather than an esm one imported using `import`

docs/cli.md
davidjgoss: CommonJS files can be `import()`ed fine. The reason we're avoiding it with `.js` is because the user might be relying on some transpilation and we don't support that yet with `import()`. To be honest I think this just adds confusion so I'll remove `.cjs` from the default handling, users can always specify it for either `require` or `import`.

docs/esm.md
davidjgoss: I would say yes.

docs/cli.md
aurelien-reeves: Ok, let's do that

PR #1932, opened 2022-02-24
Replace contructor parameter object with regular arguments

mattwynne:
### 🤔 What's changed?

* Replaced the parameter object for `Runtime`'s constructor with a regular list of method arguments
* Used constructor assignment
* Removed the `INewRuntimeOptions`

Working on #1770 with @bluemona was the first time I've really delved into the Cucumber-JS codebase, and the idioms around using constructor parameter objects felt like a hangover from when the codebase was written in JavaScript.

Personally, I think that unless the parameter object has some kind of meaning in the domain, it's better not to have a type for it, and to just pass a list of parameters instead.

I like the fact that we can reduce the boilerplate noise by using TypeScript's constructor assignment feature.

So this is a proposal, for discussion. I think it might be helpful to lighten the codebase a little and reduce the number of types we have to define if we follow this convention more generally. I'd be happy to potter away sending more PRs to do this in other places.

I guess this would end up being a change to the public API in some cases, and therefore might need a major release, so could be a bit disruptive?

### 📋 Checklist:

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,11 +6,7 @@ import * as messages from '@cucumber/messages'
 export { default as Cli } from './cli'
 export { parseGherkinMessageStream } from './cli/helpers'
 export { default as PickleFilter } from './pickle_filter'
-export {
-  default as Runtime,
-  INewRuntimeOptions,
-  IRuntimeOptions,
-} from './runtime'
+export { default as Runtime, IRuntimeOptions } from './runtime'
 export { default as supportCodeLibraryBuilder } from './support_code_library_builder'
 export { default as DataTable } from './models/data_table'
 export { default as TestCaseHookDefinition } from './models/test_case_hook_definition'
--- a/src/run/runtime.ts
+++ b/src/run/runtime.ts
@@ -49,12 +49,12 @@ export function makeRuntime({
       numberOfWorkers: parallel,
     })
   }
-  return new Runtime({
+  return new Runtime(
     eventBroadcaster,
     eventDataCollector,
     newId,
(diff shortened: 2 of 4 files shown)


davidjgoss (commented):
Ooh, controversial!

> Personally, I think that unless the parameter object has some kind of meaning in the domain, it's better not to have a type for it, and to just pass a list of parameters instead...I like the fact that we can reduce the boilerplate noise by using TypeScript's constructor assignment feature.

I think it depends. `Runtime` isn't the best example because it's on the public API and people do depend on it (mostly other framework authors, but also some users with niche needs) and I couldn't really get behind a breaking change for purely stylistic reasons.

Options objects are in my opinion a good choice when:

- There are lots of parameters and many are optional - thus avoiding lots of `null` or `undefined` interspersed between the parameters you're actually using
- People read the consuming code and want to understand what's what - the fact that it's an object with intentionally-named keys vs some things in a certain order can make a difference e.g. `doThing(stuff, 4, false, true)` needs some digging to understand

That said, I think a lot of the internals would benefit from your approach, especially as you say with the TypeScript shorthand for class members from the ctor. The parallel `Coordinator` I noticed is pretty unweildy in this respect when working on https://github.com/cucumber/cucumber-js/pull/1931, so maybe that'd be a good one to hit with this.

mattwynne:
> Runtime isn't the best example because it's on the public API

Yes, makes sense. Is the public API surface considered to be anything exported by `src/index.ts`?

davidjgoss:
@mattwynne yep exactly

PR #1936, opened 2022-02-27
api: allow reuse of support code

### LINKED ISSUE #1894: api: allow reuse of support code library between runs in the same process
**Describe the bug**
When loading support code files, Cucumber 8 uses `import(...)` instead of `require('..')`. This is great, but without a way to reuse support library code, specified in #1711, it would break StrykerJS integration. 

_NOTE: you might already know this to be the case, but I personally would love this to be fixed before the v8 release via this API: https://github.com/cucumber/cucumber-js/issues/1711#issuecomment-872841142. Otherwise, StrykerJS will be unable to support cucumber v8_

**To Reproduce**

This code example works for `@cucumber/cucumber@7.3.2`, but not for `@cucumber/cucumber@8.0.0-rc2`.

```js
const { Cli } = require('@cucumber/cucumber');
const { DirectoryRequireCache } = require('@stryker-mutator/util');

const cli = new Cli({
  argv: ['', ''],
  cwd: process.cwd(),
  stdout: process.stdout,
  env: process.env,
});

const cache = new DirectoryRequireCache();

async function main() {
  const result = await cli.run();
  console.log(result);
  cache.record();
  cache.clear();
  const result2 = await cli.run();
  console.log(result2);
}

main().catch(console.error);
```

In this example, the `DirectoryRequireCache` class is used, which is a utility that clears all local files from the `require.cache`.

**Expected behavior**

```
.


4 scenarios (4 passed)
12 steps (12 passed)
0m00.046s (executing steps: 0m00.009s)
{ shouldExitImmediately: false, success: true }
.


4 scenarios (4 passed)
12 steps (12 passed)
0m00.044s (executing steps: 0m00.006s)
{ shouldExitImmediately: false, success: true }
```

Actual behavior:

```
.

4 scenarios (4 passed)
12 steps (12 passed)
0m00.032s (executing steps: 0m00.004s)
{
  shouldAdvertisePublish: false,
  shouldExitImmediately: false,
  success: true
}

U

Failures:

1) Scenario: easy maths # features/simple_math.feature:7
  ? Given a variable set to 1
       Undefined. Implement with the following snippet:
  [...]
```

**Proposed solution**

Implement the proposal mentioned here: https://github.com/cucumber/cucumber-js/issues/1711#issuecomment-872841142 before the v8 release.

**Additional context**

When you use `import(...)` to load a cjs module, nodejs will use the CommonJS loader initially to load that file. However, it places it in the "import cache" (for lack of a better word), which is the cache that we cannot clear. Any subsequent `import`s of that file will be reclaimed from that cache. This means that support code cannot be reloaded after it is loaded initially.

I didn't realize this myself, until jasmine v4 got released last week, which broke the StrykerJS integration. 

See for a code example demonstrating the issue https://github.com/stryker-mutator/stryker-js/issues/2922#issuecomment-1013487907

davidjgoss:
### 🤔 What's changed?

`runCucumber` function modified to accept an existing support code library as an alternative to coordinates for loading support code. We also have a `loadSupport` function if you want to load just the support code upfront, then pass that into `runCucumber`.

See `runCucumber_spec.ts` for what the usage looks like - essentially we return the support code library from the first run, and it can be passed into calls for subsequent runs.

Also added an optional `onMessage` callback arg which makes testing easier (see changes to `cck_spec.ts`) and should be useful for consumers too.

Documentation changes waiting until the `runCucumber` API is release-ready.

Fixes #1894, another step for #1711.

Making it so you can run cucumber multiple times in the same process. Without this, required/imported modules would be cached by Node and wouldn't load again in order to register the support code. 

### 📋 Checklist:


aurelien-reeves (approved):
Looks pretty good. Did not notice anything special.

Some remains cryptic to me, but love the way acceptance tests (the cck) are actually showing the improvements that the new API is bringing 🤩

nicojs:
Hype!!!! ⭐

gravityvi:
@davidjgoss when can we expect this change to appear in the cucumber release?

nicojs:
@davidjgoss I finally got a change to implement using this API this weekend. It works like a charm 👌

See the work in progress: https://github.com/stryker-mutator/stryker-js/pull/3550

davidjgoss:
@nicojs nice!! Great to see it's working out for you.

PR #1947, opened 2022-03-02
add macos to matrix, use npm 8

davidjgoss:
- Add macos to the OS matrix for building and testing
- Use npm 8


davidjgoss:
This might not be viable due to some performance issues with macos runners https://github.com/actions/virtual-environments/issues/1336#issuecomment-789546414, going to keep experimenting.

PR #1950, opened 2022-03-03
reports: add handler for published url

### LINKED ISSUE #1797: Register custom handler to intercept URL of published report
**Is your feature request related to a problem? Please describe.**

I would like to capture the URL of the published report so I can notify an external system with the report URL.

**Describe the solution you'd like**

I want to be able to register a function that will be called with the URL of the published report - something like this:

```
import { registerPublishListener } from '@cucumber/cucumber'

registerPublishListener(async function (reportUrl: string) {
   // Do something with the URL, for example send it to Slack
})
```

**Describe alternatives you've considered**

Redirecting STDERR to a file, then parse it with a shell script and send the message. That's quite a roundabout way of doing this.

**Additional context**

This is requested by a user of Cucumber Reports. See similar request here: https://github.com/cucumber/cucumber-jvm/issues/2324

davidjgoss:
### 🤔 What's changed?

Added `setPublishedHandler` support code function so users can do something custom with the report URL once it's done.

The function name differs from what's suggested on the issue, but I went for consistency with existing stuff e.g. `setWorldConstructor`.

Very hacky implementation which shouldn't be merged until/unless we can do it a less hacky way (probably by changing the reports API a bit).

Fixes #1797.

### 📋 Checklist:

### DIFF
--- a/src/support_code_library_builder/types.ts
+++ b/src/support_code_library_builder/types.ts
@@ -69,6 +69,8 @@ export interface IParameterTypeDefinition<T> {
   preferForRegexpMatch?: boolean
 }
 
+export type IPublishedHandler = (report: { url: string }) => void
+
 export interface IDefineSupportCodeMethods {
   defineParameterType: (options: IParameterTypeDefinition<any>) => void
   defineStep: (<WorldType = IWorld>(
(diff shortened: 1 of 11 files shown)


src/support_code_library_builder/types.ts
```diff
@@ -69,6 +69,8 @@ export interface IParameterTypeDefinition<T> {
   preferForRegexpMatch?: boolean
 }
 
+export type IPublishedHandler = (report: { url: string }) => void
```
davidjgoss, line 72: Made the argument an object with a `url` property, envisaging that someday there might be more available e.g. time to expiry etc.

davidjgoss:
Closing per recent conversations, we'll move to a plugin system to allow for this.

PR #1951, opened 2022-03-04
run: handle parse errors better

### LINKED ISSUE #1404: Getting the parse error into a variable
If the feature file is not in proper gherkin format, Cucumber shows a parse error on console and aborts the tests. Following is the code that does that (this is a snippet from cucumber > lib > cli > run.js

function exitWithError(error) {
  console.error(_verror.default.fullStack(error)); // eslint-disable-line no-console
  process.exit(1);
}

I want to capture this parse error into a variable or a file. As a work around, I have added a line to function exitWithError to write the parse error to a file for me to later read the error for reporting purposes. But this solution requires my own version of cucumber package and then use the same. But the problem is that my changes will be overwritten once cucumber package is updated. 

function exitWithError(error) {
  fs.writeFileSync("./Parse_Error", error);
  console.error(_verror.default.fullStack(error)); // eslint-disable-line no-console
  process.exit(1);
}

Please suggest.

Note: Please know that I need to handle this scenario as we don't manually create cucumber feature files but import them from Jira cucumber tests. These tests are created by stakeholders who may make some minor mistakes in these cucumber tests which may lead to the parse error. I need to capture the parse error (in a file or variable), so that I can report it back to the jira ticket as a comment or something

davidjgoss:
### 🤔 What's changed?

Don't throw immediately when a parse error is encountered by gherkin - instead ensure all the errors (as well as any valid sources) are emitted and do a normal cleanup and exit.

Avoided making this part of `parseGherkinMessageStream` because that's on the public API (for now) and I didn't want to make a breaking change to its signature unnecessarily.

Fixes #1404.

Previously we'd throw on the first parse error and formatters wouldn't have a chance to do anything with it before the process exited.

### 📋 Checklist:

### DIFF
--- a/features/gherkin_parse_failure.feature
+++ b/features/gherkin_parse_failure.feature
@@ -4,17 +4,42 @@ Feature: Gherkin parse failure
   I want an error message that points me to the file
   So that I can quickly fix the issue and move on
 
-  Scenario:
+  Scenario: URI, line and column are called out in output
     Given a file named "features/a.feature" with:
       """
-      Feature: a feature name
-        Scenario: a scenario name
+      Feature: a feature name a
+        Scenario: a scenario name b
           Given a step
           Parse Error
       """
-    When I run cucumber-js
+    When I run cucumber-js with `--format message`
     Then it fails
     And the error output contains the text:
       """
       Parse error in "features/a.feature" (4:5)
       """
+
+  Scenario: All parseable sources and all parse errors are emitted
+    Given a file named "features/a.feature" with:
+      """
+      Parse Error
+      """
+    Given a file named "features/b.feature" with:
+      """
+      Feature: a feature name b
+        Scenario: a scenario name b
+          Given a step
+      """
+    Given a file named "features/c.feature" with:
+      """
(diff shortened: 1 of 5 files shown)


features/gherkin_parse_failure.feature
```diff
@@ -16,5 +16,30 @@ Feature: Gherkin parse failure
...
+        Scenario: a scenario name
+          Given a step
+      """
+    Given a file named "features/c.feature" with:
+      """
+      Parse Error
+      """
+    When I run cucumber-js with `--format message`
+    Then it fails
+    And the output contains these messages:
```
aurelien-reeves, line 39: I would just rename the step to make it ckear we check types and quantity of messages rather than the messages

aurelien-reeves (approved):
Looks good 👌

Just a small request to rename the new step definition, but this is an approval 💪

features/gherkin_parse_failure.feature
davidjgoss: Yep good point, updated now.

PR #1952, opened 2022-03-09
config: support configuration as objects

### LINKED ISSUE #751: update profile configuration file

davidjgoss:
### 🤔 What's changed?

- [x] Support objects as well as argv-style strings in configuration files
  - [x] Check them against a schema when loading (kinda the equivalent of what commander does with the argv)
  - [x] Validate the final amalgamated configuration against rules e.g. `retryTagFilter` with `retry`
- [x] Support configuration files being ESM or JSON
  - [x] Check for `cucumber.mjs` and `cucumber.json` by default
- [x] Provide a `loadConfiguration` function for the new JavaScript API, use it internally and in tests
- [x] Rationalise discrete modules for `configuration`, `api`, and `cli` without circular dependencies

This is a big diff, much of it comes from shifting the source of truth for configuration away from the argv/commander stuff inwards a bit, since we can now have configuration coming in object form as well. Also lots of documentation changes!

#### Out of scope

- Support for YAML and environment variables as other configuration mechanisms - we'll likely do these soon though!
- Adding an entry point and documentation for the new API - will be covered in the next PR

- Fixes #751 - solves the long-standing problem of the awkward argv-based configuration file format, but in a backwards-compatible way rather than a whole new concept
- Helps with #1711 - provides an easy function for API consumers to get the configuration from their user's project

A lot of documentation changes. The intent is to stop everything being talked about in terms of the CLI, and instead cite both config files and CLI as ways to influence things - hence `cli.md` getting a lot shorter and `configuration.md` being added. Also added docs for some things that weren't covered previously.

### 📋 Checklist:


src/api/runtime.ts
```diff
@@ -32,13 +29,8 @@ export function makeRuntime({
   requireModules: string[]
   requirePaths: string[]
   importPaths: string[]
-  options: Partial<IRuntimeOptions> & { parallel?: number }
+  options: IRunOptionsRuntime
 }): IRuntime {
-  // sprinkle specified runtime options over the defaults
```
davidjgoss, line 37: No longer needed here because defaulting is done as part of loading/merging/validating configuration.

features/profiles.feature
```diff
@@ -95,3 +98,22 @@ Feature: default command line arguments
   Scenario: specifying a configuration file that doesn't exist
     When I run cucumber-js with `--config doesntexist.js`
     Then it fails
+
+  Scenario: using a JSON file
```
davidjgoss, line 102: No real change needed for this to work, because `require()` works for JSON anyway.

cucumber.json
```diff
@@ -0,0 +1,20 @@
+{
```
davidjgoss, line 1: A good example of what configuration can look like now.

src/api/convert_configuration.ts
```diff
@@ -0,0 +1,81 @@
...
+      parallel: flatConfiguration.parallel,
+      retry: flatConfiguration.retry,
+      retryTagFilter: flatConfiguration.retryTagFilter,
+      strict: flatConfiguration.strict,
+      worldParameters: flatConfiguration.worldParameters,
+    },
+    formats: {
+      stdout:
+        [...flatConfiguration.format]
+          .reverse()
```
davidjgoss, line 34: `reverse` because the _last_ one should win if there's several.

src/api/index.ts
```diff
@@ -1,3 +1,8 @@
-export * from './loadSupport'
-export * from './runCucumber'
+/*
+Anything exported here will be publicly available via `@cucumber/cucumber/api`
```
davidjgoss, line 2: The entry point and documentation for the API will be in a follow-up PR.

src/configuration/check_schema.ts
```diff
@@ -0,0 +1,35 @@
+import * as yup from 'yup'
```
davidjgoss, line 1: This is essentially doing for object-based configuration what commander does for argv-based configuration - just checks it's the right shape.

src/configuration/default_configuration.ts
```diff
@@ -0,0 +1,25 @@
+import { IConfiguration } from './types'
+
+export const DEFAULT_CONFIGURATION: IConfiguration = {
```
davidjgoss, line 3: This has been effectively refactored out of `ArgvParser`, because that's no longer the only way we get config.

src/configuration/merge_configurations.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (ADDITIVE_ARRAYS.includes(key)) {
+    return mergeArrays(objValue, srcValue)
+  }
+  if (TAG_EXPRESSIONS.includes(key)) {
+    return mergeTagExpressions(objValue, srcValue)
+  }
+  return undefined
+}
+
+export function mergeConfigurations<T = Partial<IConfiguration>>(
```
davidjgoss, line 45: This does for object-based configuration what the various helpers on `ArgvParser` do for argv-based configuration.

src/configuration/types.ts
```diff
@@ -1,41 +1,26 @@
...
-  stdout?: string
-  files?: Record<string, string>
-  publish?:
-    | {
-        url?: string
-        token?: string
-      }
-    | false
-  options?: IParsedArgvFormatOptions
+export interface IConfiguration {
```
davidjgoss, line 4: The set of options supported in both files and CLI.

src/configuration/validate_configuration.ts
```diff
@@ -0,0 +1,9 @@
+import { IConfiguration } from './types'
+
+export function validateConfiguration(configuration: IConfiguration): void {
```
davidjgoss, line 3: Once all configuration is loaded and merged together, here we validate if any options are contradictory - schema conformance is already checked earlier.

src/runtime/index.ts
```diff
@@ -33,16 +33,6 @@ export interface IRuntimeOptions {
   worldParameters: any
 }
 
-export const DEFAULT_RUNTIME_OPTIONS: IRuntimeOptions = {
```
davidjgoss, line 36: As above, defaulting now handled as part of configuration load/merge/validate.

docs/cli.md
```diff
@@ -1,178 +1,59 @@
...
-### Format Options
+```shell
+$ cucumber-js --help
+```
 
-You can pass in format options with `--format-options <JSON>`.
+## Exiting
 
-See [Formatters](./formatters.md).
+By default, cucumber exits when the event loop drains. Use the `exit` configuration option in order to force shutdown of the event loop when the test run has finished:
```
aurelien-reeves, line 50: Not related to the current PR: maybe we could rename that `exit` option to something else that self-explained that this is kinda `workaround`, like `forceExit`, or `forceShutdown` which is even more cryptic, so users may understand what they are doing when using it because they would have actually read the doc?

davidjgoss: Yeah this is a good point - perhaps we change to `forceExit` (file) and `--force-exit` and maintain `--exit` for backwards compatibility?

src/configuration/default_configuration.ts
```diff
@@ -0,0 +1,25 @@
...
+  backtrace: false,
+  dryRun: false,
+  exit: false,
+  failFast: false,
+  format: [],
```
aurelien-reeves, line 8: I noticed earlier that we now default to the progress formatter?
Could this be part of this default configuration object rather than in code?

Or I may be wrong?

src/configuration/from_file_spec.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    try {
+      await fromFile(cwd, 'cucumber.js', ['nope'])
+      expect.fail('should have thrown')
+    } catch (error) {
+      expect(error.message).to.eq(`Requested profile "nope" doesn't exist`)
+    }
+  })
+
+  it('should get single profile config from file', async () => {
```
aurelien-reeves, line 48: Would it be possible to add some checks with other types of file, like json, or esm?
Or do we rely on acceptance tests for that?

aurelien-reeves (commented):
Great work, on both options and documentation! 🤩
A few comments, but it looks really great, well done (as usual 😅)

docs/cli.md
aurelien-reeves: Good idea to keep it for backward compatibility, but not mentioning it anymore in any documentation?
Is it possible with commander to do something like this: `--[force-]exit`?

docs/dry_run.md
```diff
@@ -1,12 +1,11 @@
...
+You can run cucumber-js in "Dry Run" mode:
 
-```shell
-$ cucumber-js --dry-run
-```
+- In a configuration file `{ dryRun: true }`
+- On the CLI `$ cucumber-js --dry-run`
 
-The effect is that cucumber-js will still do all the aggregation work of looking at your feature files, loading your support code etc but without actually executing the tests. Specifically:
+The effect is that Cucumber will still do all the aggregation work of looking at your feature files, loading your support code etc but without actually executing the tests. Specifically:
```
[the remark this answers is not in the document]
davidjgoss: Isn't this covered later in the same doc?

docs/configuration.md
```diff
@@ -0,0 +1,123 @@
...
+By default, Cucumber finds features that match this glob (relative to your project's root directory):
+
+```
+features/**/*.{feature,feature.md}
+```
+
+If your features are somewhere else, you can override this by proving your own [glob](https://github.com/isaacs/node-glob) or directory:
+
+- In a configuration file `{ paths: ['somewhere-else/**/*.feature'] }`
+- On the CLI `$ cucumber-js somewhere-else/**/*.feature`
```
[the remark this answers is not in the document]
davidjgoss: > With the CLI, the path keyword is optional?

There's no `--paths` option on the CLI as they're positional arguments. To support config as objects we needed to use a property name so we have `paths`.

To repeat on the CLI it's just

```shell
$ cucumber-js features/**/*.feature acceptance/**/*.feature.md
```

src/api/types.ts
```diff
@@ -1,5 +1,47 @@
...
+  order: PickleOrder
+}
+
+export type IRunOptionsRuntime = IRuntimeOptions & { parallel: number }
+
+export interface IRunOptionsFormats {
+  stdout: string
+  files: Record<string, string>
+  publish:
+    | {
```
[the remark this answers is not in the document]
davidjgoss: Prettier formats it this way when there's a [union type](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types) spanning multiple lines. I guess it's kinda similar to the trailing comma with object literals.

src/cli/helpers.ts
```diff
@@ -262,3 +238,10 @@ export function emitSupportCodeMessages({
   emitTestCaseHooks(supportCodeLibrary, eventBroadcaster)
   emitTestRunHooks(supportCodeLibrary, eventBroadcaster)
 }
+
+export function isTruthyString(s: string | undefined): boolean {
```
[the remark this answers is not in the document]
davidjgoss: Good point, will refactor.

src/configuration/from_file_spec.ts
davidjgoss: Fair, I'll add some more tests here.

src/configuration/default_configuration.ts
davidjgoss: We always defaulted to the progress formatter, it just happened elsewhere in the code before. We could add it as a default and have any user-specified ones merged in the normal way but it would be a bit strange. Say the user specifies `--format summary`, the resolved configuration would have `format: ['progress', 'summary']`. The latter one would still take precedence, but still feels strange. What do you think?

docs/dry_run.md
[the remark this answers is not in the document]
aurelien-reeves: My bad 😅
This was hidden in the github diff tool

src/api/types.ts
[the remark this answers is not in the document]
aurelien-reeves: Ok, thanks

src/configuration/default_configuration.ts
aurelien-reeves: Oh, ok. Thanks for the explanation 👌

docs/configuration.md
```diff
@@ -0,0 +1,123 @@
...
+And the same in ESM format:
+
+```js
+export default {
+  parallel: 2,
+  format: ['html:cucumber-report.html']
+}
+```
+
+And the same in CommonJS format:
```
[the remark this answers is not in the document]
davidjgoss: Thanks!

PR #1955, opened 2022-03-16
api: clean up, add docs, add entry point

davidjgoss:
### 🤔 What's changed?

- Clean up some details of the new API, including making arguments optional where possible
- Add API Extractor to analyse and generate docs
  - Tweak some code to make this come out better - consistent naming of types/interfaces, split out some named interfaces for things that were just inline before, etc - e.g. `IRunEnvironment` and `mergeEnvironment` is all about that
  - Eventually I'd like to also document the main entry point with this, but that's a huge piece of work and it seems like a good idea to trial it on a smaller scale here
- Add an entry point so you can require or import from `@cucumber/cucumber/api` - keen to avoid adding yet more stuff to the main entry point, which should be kept for normal end user functions

https://github.com/cucumber/cucumber-js/issues/1711

### 📋 Checklist:


davidjgoss:
When reviewing, bear in mind the following things are generated:

- `reports/cucumber.api.md`
- `docs/api/**`

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
   ([#1136](https://github.com/cucumber/cucumber-js/issues/1136)
   [#1721](https://github.com/cucumber/cucumber-js/pull/1721))
 - Support for configuration to be objects instead of argv strings, and for configuration files in ESM and JSON formats ([#1952](https://github.com/cucumber/cucumber-js/pull/1952))
+- New API for running Cucumber programmatically ([#1955](https://github.com/cucumber/cucumber-js/pull/1955))
```
aurelien-reeves, line 21: This is a big thing. I would introduce it a little bit, and at least add a link pointing to the api documentation.

docs/javascript_api.md
```diff
@@ -0,0 +1,37 @@
...
+## Complex example
+
+```javascript
+import { loadConfiguration, loadSupport, runCucumber } from '@cucumber/cucumber/api'
+
+export async function runTests(directory, configFile, failFast) {
+  // things we need to specify about the environment
+  const environment = { cwd: directory }
+  // load configuration from a particular file, and override a specific option
+  const { runnable } = await loadConfiguration({ file: configFile, provided: { failFast } }, environment)
```
aurelien-reeves, line 30: What does `runnable` mean?
(I guess I know what it means - I put myself as a user here ;))

reports/cucumber.api.md
```diff
@@ -0,0 +1,144 @@
+## API Report File for "@cucumber/cucumber"
```
aurelien-reeves, line 1: What is the purpose of that file?

aurelien-reeves (commented):
I did not review everything yet
I still have a few comments. No big deal :)

CHANGELOG.md
davidjgoss: Yes good point. We'll do a thorough release note for 8.0.0 but this is a bit terse.

reports/cucumber.api.md
davidjgoss: It's basically the source of truth for API Extractor. When it runs, it generates a fresh version of this in a staging file, then compares to this and checks if anything changed. So:

- If you unwittingly make a change to the API surface, it fails in CI
- When you run `npm run docs:local`, it overwrites this, and then the diff on the PR gives a nice summary of changes for the reviewer.

docs/javascript_api.md
davidjgoss: Right, so on the returned configuration object we have:

- `runnable` - the structured configuration object that can be passed to `runCucumber` (and can have the `support` field swapped out for an existing support code library)
- `original` the "flat" configuration that we've pulled together from a file (profiles) and whatever overrides were passed in - useful for debugging, and also has a couple of properties we don't include on `runnable` since they aren't used except in the CLI (`forceExit` and `publishQuiet`)

Maybe the fact I'm destructuring in the examples makes it confusing? It could change to like

```javascript
const configuration = await loadConfiguration()

const { success } = await runCucumber(configuration.runnable)
```

But also yeah, open to better names without being overly verbose. It's not too late (yet).

reports/cucumber.api.md
aurelien-reeves: Nice! Thanks for the explanation

docs/javascript_api.md
aurelien-reeves: Dunno if you saw my last edit

Yeah, I noticed the output of the loadConfiguration API.

For the documentation, indeed your snippet may be more clear.
But still, "runnable" for something that is actually not runnable (it does not have a "run" method)

What about one of the following, or something similar, to replace `originalConfiguration`
- `options`
- `userConfiguration`
- `requestedConfiguration`

and for `runnable`:
- `configuration`
- `cucumberConfiguration`
- `sanitizedConfiguration`

?

davidjgoss: I think `userConfiguration` and `runConfiguration` is a winner, I'll make that edit - thanks!

davidjgoss: (this is done now)

docs/javascript_api.md
```diff
@@ -0,0 +1,37 @@
...
+  // load the support code upfront
+  const support = await loadSupport(runnable, environment)
+  // run cucumber, using the support code we loaded already
+  const { success } = await runCucumber({ ...runnable, support }, environment)
```
aurelien-reeves:
```suggestion
  const { runConfiguration } = await loadConfiguration({ file: configFile, provided: { failFast } }, environment)
  // load the support code upfront
  const support = await loadSupport(runnable, environment)
  // run cucumber, using the support code we loaded already
  const { success } = await runCucumber({ ...runConfiguration, support }, environment)
```

docs/javascript_api.md
davidjgoss: Whoops! Fixed now :)

aurelien-reeves (approved):
Looks pretty good.

I did not review the whole API documentation, still I approve how it is generated and the improvements regarding the API itself

docs/javascript_api.md
aurelien-reeves: :D

docs/javascript_api.md
```diff
@@ -0,0 +1,37 @@
...
+
+export async function runTests(directory, configFile, failFast) {
+  // things we need to specify about the environment
+  const environment = { cwd: directory }
+  // load configuration from a particular file, and override a specific option
+  const { runConfiguration } = await loadConfiguration({ file: configFile, provided: { failFast } }, environment)
+  // load the support code upfront
+  const support = await loadSupport(runnable, environment)
+  // run cucumber, using the support code we loaded already
+  const { success } = await runCucumber({ ...runnable, support }, environment)
```
aurelien-reeves, line 34:
```suggestion
  const { success } = await runCucumber({ ...runConfiguration, support }, environment)
```

aurelien-reeves (changes requested):

aurelien-reeves (approved):

docs/javascript_api.md
mattwynne: Could `runConfiguration` be called a `testSuite`?

PR #1959, opened 2022-03-21
cli: warn when untested node version

davidjgoss:
### 🤔 What's changed?

Distinguish between supported (error if not met) and tested (warning if not met) Node.js versions.

See https://github.com/cucumber/cucumber-js/pull/1958#issuecomment-1073451670

- :bank: Refactoring/debt/DX (improvement to code design, tooling, docuemntation etc. without changing behaviour)

### 📋 Checklist:


package.json
```diff
@@ -188,6 +188,9 @@
   "engines": {
     "node": "12 || 14 || >=16"
   },
+  "enginesTested": {
```
davidjgoss, line 191: Not a standard field but I think it's worth keeping the two things together.

src/cli/validate_node_engine_version_spec.ts
```diff
@@ -3,66 +3,69 @@ import * as sinon from 'sinon'
 import { validateNodeEngineVersion } from './validate_node_engine_version'
 
 describe(validateNodeEngineVersion.name, () => {
+  const readPackageJSON = () => ({
```
davidjgoss, line 6: Thoughts on just dropping this and having the test read the package.json as normal? It'd mean updating the test when we change the supported/tested versions but I don't think that's necessarily a bad thing. Also would potentially catch mistakes with updates to the package.json.

aurelien-reeves (approved):

src/cli/validate_node_engine_version_spec.ts
mattwynne: 🤷 the original idea was to allow us to keep the examples in the tests more focused/simple and perhaps cover edge cases that aren't in the `package.json` today, although it looks like they've evolved to pretty much mirror what's in `package.json` now. I personally prefer having my tests tell the complete story in one file, otherwise you can end up in a "[magic number](https://en.wikipedia.org/wiki/Magic_number_(programming))" situation.

Still, if you want to be able to test what's actually in the `package.json`, maybe that deserves a specific test or two as well?

PR #1962, opened 2022-03-23
docs: UPGRADING.md for 8.0.0, some tweaks and tidy up CHANGELOG.md

davidjgoss:
Adds instructions on dealing with the breaking changes. Also renames and reformats the doc to `UPGRADING.md` per standard defined in <https://github.com/cucumber/.github/blob/main/RELEASING.md#major-release>.

Final bug fix [here](https://github.com/cucumber/cucumber-js/pull/1963), and then with this we are ready for 8.0.0


aurelien-reeves (commented):
Great!

Sorry for the delay in the review :(

PR #1963, opened 2022-03-24
fix: ensure absolute paths (as file urls) can be loaded for formatters, snippets

### LINKED ISSUE #1920: Windows path with the --format option works in v7, is broken in v8
Testing version 8.0.0-rc.2 running directly from Intellij IDEA via the Cucumber.js plugin. It generates a command line such as this:

C:\Dev\Apps\nodejs\node.exe C:\Dev\ws\sandbox\NLRTestAutomation\node_modules\@cucumber\cucumber\bin\cucumber-js C:\Dev\ws\sandbox\NLRTestAutomation\features\search-plan.feature --tags @env-shared --format c:\Dev\Home\ideau\plugins\CucumberJavaScript\lib\cucumberjs_formatter_v7.js --format-options "{\"cucumberLibPath\": \"C:/Dev/ws/sandbox/NLRTestAutomation/node_modules/@cucumber/cucumber/lib\"}" --format-options "{\"colorsEnabled\": true}" --publish-quiet --require C:/Dev/ws/sandbox/NLRTestAutomation/features

It immediately fails with this error:

Error: EISDIR: illegal operation on a directory, open 'C:\Dev\ws\sandbox\NLRTestAutomation'

**Expected behavior**

Running the same with version 7.3.2 worked fine.

**Desktop**
 - OS: Windows 10
 - Browser: n/a
 - Version: n/a

**Additional context**

It appears the cause is having the Windows drive letter in the path. If I change this:

```
--format c:\Dev\Home\ideau\plugins\CucumberJavaScript\lib\cucumberjs_formatter_v7.js
```

To this:

```
--format \Dev\Home\ideau\plugins\CucumberJavaScript\lib\cucumberjs_formatter_v7.js
```

It works fine.

davidjgoss:
### 🤔 What's changed?

As part of the ESM support, we started loading formatters and snippet syntaxes with `await import()`. Unlike `require()`, paths you pass to `import()` have to be a valid URL, and for absolute paths that means using the `file://` scheme.

This PR:

- Adds a test to prove that classes can be successfully loaded based on a `file://` URL
- Fixes some code in the configuration module in order that `file://` URLs are correctly accounted for when parsing the `--format` option (which unfortunately uses `:` as the delimiter)
- Adds a note to the documentation on formatters.

This will also be called out in the migration guide via https://github.com/cucumber/cucumber-js/pull/1962

Fixes #1920.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 
 ## [Unreleased]
 ### Fixed
-- Allow `file://` URLs to be used as formatter/snippet paths in options
+- Allow `file://` URLs to be used as formatter/snippet paths in options ([#1963](https://github.com/cucumber/cucumber-js/pull/1963) [#1920](https://github.com/cucumber/cucumber-js/issues/1920))
 
 ## [8.0.0-rc.3] - 2022-03-21
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
+### Fixed
+- Allow `file://` URLs to be used as formatter/snippet paths in options
 
 ## [8.0.0-rc.3] - 2022-03-21
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ See [docs/profiles.md](./docs/profiles.md#using-another-file-than-cucumberjs) fo
 
 ### Changed
 - Relative paths for custom snippet syntaxes must begin with `.` ([#1640](https://github.com/cucumber/cucumber-js/issues/1640))
+- Absolute paths for custom formatters and snippet syntaxes must be a valid `file://` URL
 - Use performance timers for test case duration measurement.
 [#1793](https://github.com/cucumber/cucumber-js/pull/1793)
 
--- a/docs/formatters.md
+++ b/docs/formatters.md
@@ -14,6 +14,7 @@ For each value you provide, `TYPE` should be one of:
 * The name of one of the built-in formatters (below) e.g. `progress`
 * A module/package name e.g. `@cucumber/pretty-formatter`
 * A relative path to a local formatter implementation e.g. `./my-customer-formatter.js`
+* An absolute path to a local formatter implementation in the form of a `file://` URL
 
 If `PATH` is supplied, the formatter prints to the given file, otherwise it prints to `stdout`.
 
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
+### Fixed
```
aurelien-reeves, line 11: It looks more like a breaking change than a fix

I see in option_splitter_spec that it seems absolute windows paths won't be working anymore if not prefixed with `file://`?

If I am mistaken, could it be possible to keep a test in option_splitter_spec that make sure absolute windows path are still working properly?

davidjgoss: > It looks more like a breaking change than a fix...

Yeah, sort of - the breaking change actually came in `8.0.0-rc.1` when we did the initial ESM support work and changed to using `import()` for loading custom classes. Unfortunately the option splitter didn't have awareness of `file://` so it was treating the colon as a delimiter, making it impossible to actually use a `file://` URL - once a valid `file://` URL makes it through to `FormatterBuilder`, it works as intended.

Happy to recast this as a breaking change though - or perhaps retroactively add a breaking change entry to the rc.1 changelog. What do you think?

> I see in option_splitter_spec that it seems absolute windows paths won't be working anymore if not prefixed with file://?

Yes, and same for unix absolute paths. We could try to preserve backwards compatibility with some pattern matching and then prepend `file://` internally, but honestly it feels cleaner and clearer to just say "absolute paths must be a valid file URL". It's definitely an edge case - I'd expect the majority of formatters to be within a file or package local to the project.

aurelien-reeves: > > It looks more like a breaking change than a fix...
> 
> Yeah, sort of - the breaking change actually came in `8.0.0-rc.1` when we did the initial ESM support work and changed to using `import()` for loading custom classes. Unfortunately the option splitter didn't have awareness of `file://` so it was treating the colon as a delimiter, making it impossible to actually use a `file://` URL - once a valid `file://` URL makes it through to `FormatterBuilder`, it works as intended.
> 
> Happy to recast this as a breaking change though - or perhaps retroactively add a breaking change entry to the rc.1 changelog. What do you think?
> 

Yeah, maybe adding it retroactively to rc.1 would do the trick :+1: 

> > I see in option_splitter_spec that it seems absolute windows paths won't be working anymore if not prefixed with file://?
> 
> Yes, and same for unix absolute paths. Again this was the case since `8.0.0-rc.1`. We could try to preserve backwards compatibility with some pattern matching and then prepend `file://` internally, but honestly it feels cleaner and clearer to just say "absolute paths must be a valid file URL". It's definitely an edge case - I'd expect the majority of formatters to be within a file or package local to the project.

I am okay with using "file://" only. It was just to be sure to have a full understanding of the changes

davidjgoss: Thanks, done that changelog update.

aurelien-reeves (approved):

PR #1985, opened 2022-04-04
fix: try require first for custom formatters so transpiled ones work

davidjgoss:
### 🤔 What's changed?

Tweak the loading of custom formatters and snippet syntaxes so we try `require()` first and then try `import()` if it turns out the target file is a module (ESM).

This mirrors how we are also now loading the configuration file.

https://cucumberbdd.slack.com/archives/C6QJ6N695/p1649071066867509

An edge case arguably, but some users in the wild have formatters in their project written in TypeScript (or something else that transpiles) and take advantage of the fact that `--require-module` means their transpiler has registered by the time the formatter is loaded.

This is quite cool and something we should avoid breaking if possible. We broke it with earlier RCs because we switched to exclusively using `import()` to load custom formatters, which doesn't yet account for just-in-time transpiled code.

In 1 or 2 major releases (when loader hooks are a solved problem) we'll ditch the `require()` bit and just use `import()` which will simplify things a lot but for now this works reliably and is well covered by tests.

### 📋 Checklist:

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CO
 ## [Unreleased]
 ### Fixed
 - Allow `file://` URLs to be used as formatter/snippet paths in options ([#1963](https://github.com/cucumber/cucumber-js/pull/1963) [#1920](https://github.com/cucumber/cucumber-js/issues/1920))
+- Allow custom formatters to rely on `--require-module` transpilers ([#1985](https://github.com/cucumber/cucumber-js/pull/1985))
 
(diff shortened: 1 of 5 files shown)


aurelien-reeves (commented):
Looks nice!

Would it be hard to add a test (maybe a scenario?) that would help to make sure we were able to reproduce the error before the patch, and then validate the patch?

davidjgoss:
@aurelien-reeves thanks - several test in `builder_spec.ts` would fail without the code change here. I'll have a look at the feature tests too.

ludeknovy (approved):
Thanks for the fix ❤️

aurelien-reeves (approved):
🤩

Issue #1551, opened 2021-01-25
--format-options '{"colorsEnabled": false}' does not apply on error messages


hdorgeval:
Hello Cucumber Team,

Yet another issue related to displaying error messages with the internal HTML formatter:

Given I disable colors on the CLI with `--format-options '{"colorsEnabled": false}'`
When a step fails
Then all logs are uncolored
But the Error message still uses `Chalk` red/green colors
Example in the console:
![Screen Shot 2021-01-25 at 12 04 28](https://user-images.githubusercontent.com/9698924/105697981-8ec9d800-5f05-11eb-96b9-8a27076d729f.png)
Example in the html report
![Screen Shot 2021-01-25 at 12 29 25](https://user-images.githubusercontent.com/9698924/105700351-0a795400-5f09-11eb-9c7f-f19b016dfd27.png)

And the corresponding envelope is:
```json
{
  "testStepFinished": {
    "testStepResult": {
      "status": "FAILED",
      "message": "Error: \u001b[2mexpect(\u001b[22m\u001b[31mreceived\u001b[39m\u001b[2m).\u001b[22mtoBe\u001b[2m(\u001b[22m\u001b[32mexpected\u001b[39m\u001b[2m) // Object.is equality\u001b[22m\n\nExpected: \u001b[32m13\u001b[39m\nReceived: \u001b[31m12\u001b[39m\n    at CustomWorld.<anonymous> (/Users/HDO/VSCodeProjects/cucumber7-ts-starter/step-definitions/maths/simple-maths-steps.ts:19:34)",
      "duration": {
        "seconds": "0",
        "nanos": 12000000
      }
    },
    "timestamp": {
      "seconds": "1611571916",
      "nanos": 62000000
    },
    "testStepId": "71ddfe75-49af-4db7-b5ba-d264dfd2fd64",
    "testCaseStartedId": "75fd16c7-5a2a-45b2-8d7e-907f9dc03988"
  }
```

In v6 the work around was to set `--no-color` ( see [chalk support color](https://github.com/chalk/chalk#chalksupportscolor)) on the CLI.

Since it not possible any more in v7, the only way to fix this issue in v7 is to set the environment variable: 
```js
process.env['FORCE_COLOR'] = '0';
```

It would be great if you could either give back support for the `--no-color` option on the CLI or that you set the `FORCE_COLOR` environment variable.

In resume:
Given I disable colors on the CLI with `--format-options '{"colorsEnabled": false}'`
[...]

PR #2026, opened 2022-05-01
format: fix issues with colors support

### LINKED ISSUE #2011: formatter: colorsEnabled: true not honoured
### 👓 What did you see?

When outputting to a non-TTY environment, using `colorsEnabled: true` in the format options doesn't result in colours being output.

Colours to be output, because of this logic:

```ts
if (doesNotHaveValue(configuration.options.colorsEnabled)) {
      typeOptions.parsedArgvOptions.colorsEnabled = (
        stream as TtyWriteStream
      ).isTTY
    }
```

In other words, if we specify the option, it should output colours without checking if the stream is TTY.

cucumber-js 8.1.2

TBA

davidjgoss:
### 🤔 What's changed?

We recently switched from `colors` to `chalk` for handling terminal colors. `chalk` includes its own rather comprehensive support checks at runtime, which had the unintended effect of disregarding a user-provided `colorsEnabled: true` (fixes #2011).

Now, rather than our own support check (which was just `stream.isTTY`), we explicitly delegate to chalk's detection library, for the specific output stream in question, and then construct our own chalk instance with the result (having overrided it based on the format option if necessary). See the feature file and docs for more detail, but the upshot is:

- Auto detection of colors support is better than it was
- The `colorsEnabled` option now works again
- Users can also use `FORCE_COLOR` which chalk supports, and we recommend it (for reasons in the docs)

🚨 After considering this, I'd like to go a bit further and deprecate (with intent to eventually remove) the `colorsEnabled` format option in favour of having users just use the kinda-standard `FORCE_COLOR` environment variable, which can at least influence other tools like assertion libraries which use `chalk` directly (see #1551) and would remove some complexity on our side (as we'd just fully delegate the logic to `supports-color`). Mocha folks have also leant into `FORCE_COLOR` to some degree in https://github.com/mochajs/mocha/issues/3641

Unless there are some implications I'm not seeing for integrators using Cucumber programmatically - paging @nicojs @jan-molak?

### 📋 Checklist:

----


src/formatter/get_color_fns.ts
```diff
@@ -48,3 +57,15 @@ export default function getColorFns(enabled: boolean): IColorFns {
...
   }
 }
+
+function detectSupport(
+  stream: Writable,
+  env: NodeJS.ProcessEnv,
+  enabled?: boolean
+): ColorInfo {
+  const support: ColorInfo = supportsColor(stream)
+  if ('FORCE_COLOR' in env || doesNotHaveValue(enabled)) {
```
davidjgoss, line 67: If `FORCE_COLOR` is present, we'll just let the library handle that and not bother with our own option.

src/types/supports-color/index.d.ts
```diff
@@ -0,0 +1,17 @@
+declare module 'supports-color' {
```
davidjgoss, line 1: DefinitelyTyped types for this module were wrong.

docs/formatters.md
```diff
@@ -39,9 +39,17 @@ This option is repeatable, so you can use it multiple times and the objects will
 
 Some options offered by built-in formatters:
 
-- `colorsEnabled` - if set to `false`, colors in terminal output are disabled
+- `colorsEnabled` - see below
```
aurelien-reeves, line 42:
```suggestion
- `colorsEnabled` - [see below](#colored-output)
```

src/formatter/get_color_fns.ts
aurelien-reeves: Maybe that comment would be useful in the code itself?

aurelien-reeves (commented):
Looks good 🙂

Changelog entry is missing

Also, I do not get why do we add a stream to `getColorFns`?

davidjgoss:
> Also, I do not get why do we add a stream to `getColorFns`?

We pass in the stream that the formatter is outputting to, so that `supports-colors` will detect the colors support for that particular stream, otherwise it would just look at `process.stdout` which may or may not be right.

aurelien-reeves (approved):

aurelien-reeves (approved):

binarymist:
Thanks Team, tested and all good!

PR #2038, opened 2022-05-10
fix return type of step hook function to allow async

### LINKED ISSUE #2037: BeforeStep and AfterStep hooks do not allow asynchronous functions - linting (es-lint) error
### 👓 What did you see?

The cucumber-js api reference for BeforeStep and AfterStep indicates that one can use async functions as the function parameter to BeforeStep and AfterStep, just like Before/After and BeforeAll/AfterAll ([API reference](https://github.com/cucumber/cucumber-js/blob/main/docs/support_files/api_reference.md)). However, when attempting to use an async function as an argument to AfterStep, I got the following typescript linting error:
`Promise returned in function argument where a void return was expected`

After doing some poking around, of the three hook types, the `TestStepHookFunction` type only allows for a return type of void, whereas the other two types, `TestStepHookFunction` and `TestStepFunction` have the following return types: `any | Promise<any>` ([code](https://github.com/cucumber/cucumber-js/blob/3928c673378ccabf2138b7c55873f6981e9eb134/src/support_code_library_builder/types.ts#L35)).

I expected that `TestStepHookFunction`, `TestStepHookFunction`, and `TestStepFunction` all had the return type of `any | Promise<any>`.

Cucumber-js version: 7.3.2
Typescript version: 4.6.3
eslint: 8.12.0

1. Install cucumber-js version 7.3.2
2. Install and apply eslint version 8.12.0
3. Create a BeforeStep and pass in an async function, something like:
```
BeforeStep(async function(): Promise<void> {
await fetch("http://github.com");
})
```
4. Linter/compiler throws expected error: `Promise returned in function argument where a void return was expected`

### 📚 Any additional context?

----
 
*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/bug_report.md)*

davidjgoss:
### 🤔 What's changed?

Fix return type of step hook function to allow async, to be more accurate and prevent compile/lint errors for users.

Fixes #2037.

### 📋 Checklist:

----

### DIFF
--- a/src/support_code_library_builder/types.ts
+++ b/src/support_code_library_builder/types.ts
@@ -35,7 +35,7 @@ export type TestCaseHookFunction<WorldType> = (
 export type TestStepHookFunction<WorldType> = (
   this: WorldType,
   arg: ITestStepHookParameter
-) => void
+) => any | Promise<any>
 
 export type TestStepFunction<WorldType> = (
   this: WorldType,
--- a/test-d/hooks.ts
+++ b/test-d/hooks.ts
@@ -17,6 +17,14 @@ After(function () {})
 BeforeStep(function () {})
 AfterStep(function () {})
 
+// should allow hook functions to be async
+BeforeAll(async function () {})
+AfterAll(async function () {})
+Before(async function () {})
+After(async function () {})
+BeforeStep(async function () {})
+AfterStep(async function () {})
+
 // should allow typed arguments in hooks
 Before(function (param: ITestCaseHookParameter) {})
 After(function (param: ITestCaseHookParameter) {})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](https://github.com/cucumber/cucumber/blob/master/CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
+### Fixed
+- Fix return type of step hook function to allow async functions ([#2038](https://github.com/cucumber/cucumber-js/pull/2038))
 
 ## [8.2.0] - 2022-05-05
 ### Changed
--- a/features/support/hooks.ts
+++ b/features/support/hooks.ts
@@ -1,11 +1,10 @@
-import { After, Before, formatterHelpers } from '../../'
+import { After, Before, formatterHelpers, ITestCaseHookParameter } from '../../'
 import fs from 'fs'
 import fsExtra from 'fs-extra'
 import path from 'path'
 import tmp from 'tmp'
 import { doesHaveValue } from '../../src/value_checker'
 import { World } from './world'
-import { ITestCaseHookParameter } from '../../src/support_code_library_builder/types'
 import { warnUserAboutEnablingDeveloperMode } from './warn_user_about_enabling_developer_mode'
 
 const projectPath = path.join(__dirname, '..', '..')


src/support_code_library_builder/types.ts
```diff
@@ -35,7 +35,7 @@ export type TestCaseHookFunction<WorldType> = (
 export type TestStepHookFunction<WorldType> = (
   this: WorldType,
   arg: ITestStepHookParameter
-) => void
+) => any | Promise<any>
```
mattwynne, line 38: `any` implies we expect an interesting return value. Do we look at the return value of a hook? If not, I suggest `void | Promise<void>` is better.

test-d/hooks.ts
```diff
@@ -17,6 +17,14 @@ After(function () {})
 BeforeStep(function () {})
 AfterStep(function () {})
 
+// should allow hook functions to be async
+BeforeAll(async function () {})
```
mattwynne, line 21: Did you not want to use our own types here for these functions?

e.g. for `BeforeStep` and `AfterStep` could we not use the `TestStepHookFunction` from above? Or am I missing something?

davidjgoss: That type is implicitly being used - the goal with this test is to write what a user would typically write and make sure it passes the type check.

src/support_code_library_builder/types.ts
davidjgoss: Good question.

It's not explicitly documented (per https://github.com/cucumber/cucumber-js/issues/1312#issuecomment-1016807156 which we need to address) but the return value can be "skipped" or "pending" as with regular steps.

It would be good to add this explicitly to the typings but I'm wary of suddenly causing compilation issues for people that e.g. are using chain promises and implicitly returning the result.

I'll do a little experimenting.

aurelien-reeves: `string | void | Promise<string | void>` won't work?

davidjgoss: (It doesn't really make sense to `return 'pending'` from a `BeforeStep`, so we should reconsider that, which I think was just an accident of the implementation.)

davidjgoss: > `string | void | Promise<string | void>` won't work?

My thought was around user code like:

```
BeforeStep(() => {
  return somethingAsyncThatReturnsPromise()
})
```

Where they are returning the promise so cucumber awaits it, but it resolves to some value that isn't a string so it fails the compilation, and now they have to refactor. Maybe less relevant now with `async/await` being so prevalent though.

davidjgoss: > We may use `any` in a first time, implementing a deprecation notice when not being an expected return type, then fix it in a version after that?

Yeah that sounds like a good plan. We need to rationalise this, plus the return types for the other hooks and steps, which I think should happen in a major release.

mattwynne: Just FYI, you can use a concrete string value as a type in Typescript, so it could be `"skipped" | "pending" | void` which is narrower than `string | void`.

I see your point about keeping it loose though David, consider also people this:

`BeforeStep(somethingAsyncThatReturnsPromise)`

test-d/hooks.ts
mattwynne: 💡  thanks for clarifying!

src/support_code_library_builder/types.ts
davidjgoss: Yep another good example.

What do we think then to merging this and raising something new to fix all these types properly for 9.0.0?

mattwynne: Good plan. 👍

davidjgoss: New issue here https://github.com/cucumber/cucumber-js/issues/2040

PR #2089, opened 2022-07-18
rework handling and docs for invalid installations

### LINKED ISSUE #1883: false positives with install validator
**Describe the bug**
On windows 10 in git bash I execute cucumber the first time after npm install. An error is raised saying the executed install of cucumber is different from my local install, which is not correct. The trace shows that the validator compares my paths case sensitive:
`Error:
      You appear to be executing an install of cucumber (most likely a global install)
      that is different from your local install (the one required in your support files).
      For cucumber to work, you need to execute the same install that is required in your support files.
      Please execute the locally installed version to run your tests.

      Executed Path: C:\Development\web\it\analyse\src\e2e-xray\node_modules\@cucumber\cucumber\lib\index.js
      Local Path:    C:\development\web\it\analyse\src\e2e-xray\node_modules\@cucumber\cucumber\lib\index.js

    at Object.validateInstall (C:\Development\web\it\analyse\src\e2e-xray\node_modules\@cucumber\cucumber\lib\cli
\install_validator.js:28:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Cli.run (C:\Development\web\it\analyse\src\e2e-xray\node_modules\@cucumber\cucumber\lib\cli\index.js
:132:9)
    at async Object.run [as default] (C:\Development\web\it\analyse\src\e2e-xray\node_modules\@cucumber\cucumber\
lib\cli\run.js:25:18)
`

**To Reproduce**
1. Install cucumber via npm
2. Run cucumber: cucumber-js features/**/*.feature
3. Error is raised in console

**Expected behavior**
Cucumber's install_validator recognises paths "C:\Development" and "C:\development" as as equal OS paths

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: na
 - Version 10

**Additional context**

davidjgoss:
### 🤔 What's changed?

This PR improves the way we deal with invalid installations - that is, where the Cucumber instance that gets run isn't the one imported by the support code (see the new `installation.md` doc for a more expansive intro).

For global installations, we had a check early in the CLI that would throw if it looked like a global installation (one of the ways you can be invalid). However this is [prone to some false positives](https://github.com/cucumber/cucumber-js/issues/1883) depending on users' platform tooling etc. The logic has been replaced by the [is-installed-globally](https://www.npmjs.com/package/is-installed-globally) library which takes a different approach that should avoid those false positives. Also, it warns rather than throws, which is a hedge in case it has issues, but I could be talked out of that.

Before, with an invalid installation (e.g. with a duplicate dependency, or linking), you'd hit a rather cryptic error (`The "from" argument must be of type string. Received type undefined`), from which astute/lucky users could Google their way to our FAQ doc. We now intentionally detect the condition for that error (i.e. the support code library builder hasn't been initialised because it's not the instance of Cucumber that's actually running) and throw a descriptive message that links to a more useful installation doc. This will happen for global installs now too (as well as the warning).

For testing, the existing feature test remains but is tweaked. The setup of the invalid installation moves from being a tagged hook to a "Given" step - just felt right from a Gherkin perspective.

Fixes #1883.

Using a proxy to wrap all the support methods in the check. Seemed like the cleanest and most maintainable way.

### 📋 Checklist:

----


aurelien-reeves (commented):
It looks pretty good!

Regarding the installation document, in the "duplicate" paragraph, it is emphased the use of external support code. I would also emphasize the use of third parties. And I would also mention the use if an old cucumber alongside the new @cucumber/cucumber one

davidjgoss:
Great points @aurelien-reeves - I've pushed some changes to address those.

docs/installation.md
```diff
@@ -0,0 +1,76 @@
...
+
+### Deprecated package
+
+When looking at the duplicate dependency issue, it's worth checking whether anything in your project is depending on the old, deprecated `cucumber` package. Anything touching Cucumber [should be using](../UPGRADING.md#package-name) the newer `@cucumber/cucumber` package.
+
+### Linking
+
+With the shared library example above, even if you have `@cucumber/cucumber` correctly defined as a peer dependency, you can still hit the issue if you hook up the library locally using `npm link` or `yarn link` when developing or testing.
+
+This is trickier to the deal with. If you run `npm link ../my-project/node_modules/@cucumber/cucumber` from the library, this should work around it (assuming `my-project` is your host project's directory, and it's adjacent to your library in the file system).
```
aurelien-reeves, line 65:
```suggestion
This is trickier to deal with. If you run `npm link ../my-project/node_modules/@cucumber/cucumber` from the library, this should work around it (assuming `my-project` is your host project's directory, and it's adjacent to your library in the file system).
```

aurelien-reeves (approved):
Looks great :)

PR #2092, opened 2022-07-23
reimplement publish functionality as a plugin

davidjgoss:
### 🤔 What's changed?

This PR moves the "publish" functionality from being a pseudo-formatter to being a "plugin" - a new concept we're experimenting with (see https://github.com/cucumber/cucumber-js/discussions/2091).

The plugin concept stays internal here - the idea is to get feedback on this draft API and experiment with a couple more use cases before releasing anything public-facing. This PR should be mergeable any time though.

So, the design. The structure of a plugin is basically:

- It's an async function, which would be the default export for a plugin package, like with formatters now
- It takes an argument which gives it some stuff to work with - configuration, environment stuff and a way to listen for events
- It can listen for events - just Cucumber messages for now - and do stuff with them. Note that the hander is _not_ async - we don't want to be wait on message consumers sending network requests etc. This would be different for things that can affect the test run e.g. reordering pickles - we'll want to allow those to be async.
- It can optionally provide a cleanup function to be done at the end of the test run

I borrowed ideas from several places for this:

- Cypress - the `on('event', handler)` style in particular, and I also like the format of lifecycle hook naming e.g. `"filter:database"`
- ESLint - the way plugins are referenced _declaratively_ in the configuration - i.e. no writing code to hook up plugins
- React - optionally returning a "cleanup" function at the end of the initialising function is shamelessly stolen from `useEffect`

### 📋 Checklist:

----


features/publish.feature
```diff
@@ -106,7 +106,7 @@ Feature: Publish reports
   @spawn
   Scenario: when results are not published due to an error raised by the server, the banner is displayed
     When I run cucumber-js with env `CUCUMBER_PUBLISH_TOKEN=keyboardcat`
-    Then it fails
+    Then it passes
```
davidjgoss, line 109: The only functional change here. For formatters, we catch errors on the stream(s), log them and then fail the test run at the end. Publish is now not a formatter, and at any rate this felt a bit off - failing to publish the report seems like it shouldn't fail the test run? That's debatable, though.

It does raise a couple of questions about error handling though:
- If there's an uncaught error in a plugin, should we try to fail the test run gracefully-ish?
- Should there be a way for a plugin to say "The test run should be marked as a failure because XYZ"? Or should it just throw if it wants to do that?

src/api/run_cucumber.ts
```diff
@@ -25,7 +26,7 @@ export async function runCucumber(
   onMessage?: (message: Envelope) => void
 ): Promise<IRunResult> {
   const { cwd, stdout, stderr, env } = mergeEnvironment(environment)
-  const logger = new Console(stdout, stderr)
+  const logger = new Console(stderr)
```
davidjgoss, line 29: This was an oversight when initially added. This `logger` is generally used for `warn` and `error`, but even for `log` (if used) it should still go to `stderr` - only formatters get to write stuff to `stdout`.

src/api/run_cucumber.ts
```diff
@@ -47,14 +48,17 @@ export async function runCucumber(
...
   const eventBroadcaster = new EventEmitter()
   if (onMessage) {
     eventBroadcaster.on('envelope', onMessage)
   }
+  eventBroadcaster.on('envelope', (value) => plugins.emit('message', value))
   const eventDataCollector = new EventDataCollector(eventBroadcaster)
 
   let formatterStreamError = false
-  const cleanup = await initializeFormatters({
+  const cleanupFormatters = await initializeFormatters({
```
davidjgoss, line 61: Renamed for clarity.

src/formatter/index.ts
```diff
@@ -23,11 +20,7 @@ export interface FormatOptions {
   [customKey: string]: any
 }
 
-export type IFormatterStream =
-  | FsWriteStream
-  | TtyWriteStream
-  | PassThrough
-  | HttpStream
+export type IFormatterStream = Writable
```
davidjgoss, line 23: `HttpStream` was no longer applicable here but also this didn't feel like a useful type. We need a writable stream, is all.

src/plugin/types.ts
```diff
@@ -0,0 +1,20 @@
+import { IRunEnvironment, IRunOptions } from '../api'
+import { Envelope } from '@cucumber/messages'
+
+export interface PluginEvents {
```
davidjgoss, line 4: This would expand to cover other event we start to support.

mattwynne:
Thoughts, from a first pass reading over this:

1) If we go with the rule stated [above](https://github.com/cucumber/cucumber-js/pull/2092#discussion_r928166829) that "only formatters get to write stuff to `stdout`", then what happens if we embrace the idea that [formatters are just a special type of plugin](https://github.com/cucumber/cucumber-js/discussions/2091#discussioncomment-3211249)?

2) Similarly, with handling errors in the event listeners, we probably need to be consistent across plugins and formatters. For me, those event listeners should not get to impact the test run's success, but it also seems like a useful feature for errors to be reported to the user and to cause a non-zero exit status from the `cucumber` command itself.

3) I worry that the returning-a-cleanup-function thing will be non-intuitive to discover. It would feel more natural and extensible to me if there were a way to access the `SupportCodeMethods` API so you could simply add an `AfterAll` hook. What do you think about that?

4) Talking of which, if we plan to allow plugins to "Hook into core behaviour and change things" (as stated in the new docs page), I guess this use-case doesn't really give us any need to do that? So we'll want to look for other things we can factor into this plugin mechanism to help give it it's final shape, is that what you're thinking?

davidjgoss:
Thanks for the feedback @mattwynne! To your points:

1. A formatter plugin would also get a `write` function in its context argument that would enable it to write to the relevant output stream.
2. Agreed, I'll do something to catch and log. I wonder if for handlers that _change things_ (as opposed to just consuming messages) it should fail fast though?
3. Yeah, I get that it might not be the most intuitive, although it has some prior art. Allowing to add extra `AfterAll` here feels wrong to me, partly because it doesn't quite line up with what we want to be able to do (`AfterAll` hooks run on the worker not the coordinator for parallel, and are part of the test run, where there are more messages still to come) but also generally because I feel there should be a separation between user code and plugin code, it could get confusing to mix them. If you feel strongly about the return-a-cleanup-function thing, I could get on board with a `cleanup` function being part of the context argument, or maybe just `on('end', async () => {})` as a supported event. WDYT?
4. Yep exactly. I think being able to filter and reorder pickles is an obvious candidate. I'll probably make a branch from this and see where that goes.

mattwynne:
> Thanks for the feedback @mattwynne! To your points:
> 
>     1. A formatter plugin would also get a `write` function in its context argument that would enable it to write to the relevant output stream.

OK, let's cross that bridge when we come to it.

>     2. Agreed, I'll do something to catch and log. I wonder if for handlers that _change things_ (as opposed to just consuming messages) it should fail fast though?

In my view, an event protocol should be "fire and forget" from our point of view, so not a good place to let people change things. I think we should find a different mechanism for that. Again, shall we cross that bridge when we come to it?

>     3. Yeah, I get that it might not be the most intuitive, although it has some prior art. Allowing to add extra `AfterAll` here feels wrong to me, partly because it doesn't quite line up with what we want to be able to do (`AfterAll` hooks run on the worker not the coordinator for parallel, and are part of the test run, where there are more messages still to come) but also generally because I feel there should be a separation between user code and plugin code, it could get confusing to mix them. If you feel strongly about the return-a-cleanup-function thing, I could get on board with a `cleanup` function being part of the context argument, or maybe just `on('end', async () => {})` as a supported event. WDYT?

Fair enough. I hadn't thought as clearly as you about the distinction between user and plugin code, and where they would run. I still feel like there might be a need for plugins to be able to do something like what user code can do, but let's again wait until we have a use-case.

>     4. Yep exactly. I think being able to filter and reorder pickles is an obvious candidate. I'll probably make a branch from this and see where that goes.

I say we merge this in and then do that. It's good enough for now and has enough caveats on it that we can change it later.

davidjgoss:
> I say we merge this in and then do that.

Works for me! Will get the conflicts fixed and get it merged. Mind dropping an approval on here for completeness?

mattwynne (approved):

PR #2119, opened 2022-08-22
alternative stack trace approach

davidjgoss:
### 🤔 What's changed?

Stop using `stack-chain` to temporarily modify stack trace behaviour in a V8-specific manner, for when we filter stack traces (i.e. with `--backtrace` option). Pivot to parsing the stack with `error-stack-parser` without modifying how it's created, and filtering from that.

Also add some documentation around source maps.

This started with me digging around to try and understand this aspect of our codebase better, but I wanted to run with this approach and see if it's viable.

Hooking into V8's stack trace API in the way that `stack-chain` does in order to filter the stack trace for display seems like a bit of a sledgehammer to me. It also doesn't fail gracefully - in a non-V8 environment it will throw and exit, and this happens as soon as you import `stack-chain`, not just when you call methods on it. This is the only reason we can't do a live demo with StackBlitz, for example, and also seems likely to be a hindrance as we look to become less Node.js-centric and work with e.g. Deno and Bun.

On the flipside, parsing the stack trace is inevitably less scientific and more error-prone, and there will be some edge cases that go weird. For example https://github.com/stacktracejs/error-stack-parser/issues/62 which contributed to us to moving away from the stacktrace.js libraries before (I've [raised a fix](https://github.com/stacktracejs/error-stack-parser/pull/82) for this now). I think it would be worthwhile though.

This change has also fixed an issue I was seeing when creating an example TypeScript+ESM project, where source references and stack traces were coming out with wrong locations that didn't correspond to either the source or transpiled JavaScript.

This area isn't very well covered by tests, so I'm going to try and flesh some out before I mark this as ready.

- 🐛  Bug fix (non-breaking change which fixes a defect)

### 📋 Checklist:

----


src/runtime/format_error.ts
```diff
@@ -0,0 +1,22 @@
...
+      filteredStack = filterStackTrace(errorStackParser.parse(error))
+        .map((f) => f.source)
+        .join('\n')
+    } catch {
+      // if we weren't able to parse and filter, we'll settle for the original
+    }
+  }
+  return format(error, {
+    colorFns: {
+      errorStack: (stack: string) =>
```
davidjgoss, line 18: I'm kind of abusing the `assertion-error-formatter` API here, which doesn't feel great. It could maybe do with a `formatErrorStack` option where you can provide a function that rewrites the stack as desired?

package.json
```diff
@@ -309,7 +308,7 @@
     "prepublishOnly": "rm -rf lib && npm run build-local",
     "pretest-coverage": "npm run build-local",
     "pretypes-test": "npm run build-local",
-    "test-coverage": "nyc --silent mocha 'src/**/*_spec.ts' 'compatibility/**/*_spec.ts' && nyc --silent --no-clean node bin/cucumber.js && nyc report --reporter=lcov",
+    "test-coverage": "nyc --silent mocha 'src/**/*_spec.ts' 'compatibility/**/*_spec.ts' && nyc --silent --no-clean node bin/cucumber.js --tags \"not @source-mapping\" && nyc report --reporter=lcov",
```
davidjgoss, line 311: The `nyc` instrumentation seems to mess these scenarios up when we check for coverage. We could try and fix up the environment for the child process but it didn't feel worth it.

davidjgoss:
This was released in 8.8.0 https://github.com/cucumber/cucumber-js/releases/tag/v8.8.0

PR #2121, opened 2022-08-29
feat: add junit formatter

ricalbuquerque:
### 🤔 What's changed?

- Added JUnit formatter

Needed to get proper reporting in Azure Pipelines so had to develop a JUnit formatter that would not treat each step as a different test. After getting testing and started to use it I decided to make it available for the community,

- Feel free to suggest any change or possible improvement! 

### 📋 Checklist:

----

### DIFF
--- a/src/formatter/junit_formatter_spec.ts
+++ b/src/formatter/junit_formatter_spec.ts
@@ -1,0 +1,245 @@
+import { afterEach, beforeEach, describe, it } from 'mocha'
+import { expect, use } from 'chai'
+import chaiXml from 'chai-xml'
+import FakeTimers, { InstalledClock } from '@sinonjs/fake-timers'
+import timeMethods from '../time'
+import { testFormatter } from '../../test/formatter_helpers'
+import { getJsonFormatterSupportCodeLibrary } from '../../test/fixtures/json_formatter_steps'
+
+use(chaiXml)
+
+describe('JunitFormatter', () => {
+  let clock: InstalledClock
+
+  beforeEach(() => {
+    clock = FakeTimers.withGlobal(timeMethods).install()
+  })
+
+  afterEach(() => {
+    clock.uninstall()
+  })
+
+  describe('no features', () => {
+    it('outputs an empty array', async () => {
+      // Arrange
+
+      // Act
+      const output = await testFormatter({ type: 'junit' })
+
+      // Assert
+      expect(output).xml.to.deep.equal('<?xml version="1.0"?>\n<testsuites/>')
+    })
+  })
+
+  describe('one scenario with one step', () => {
+    describe('passed', () => {
+      it('outputs the feature', async () => {
+        // Arrange
+        const sources = [
+          {
+            data: [
+              'Feature: my feature',
+              '  my feature description',
+              '',
+              '  Scenario: my scenario',
+              '    my scenario description',
+              '',
+              '    Given a passing step',
+            ].join('\n'),
+            uri: 'a.feature',
+          },
+        ]
+
+        const supportCodeLibrary = getJsonFormatterSupportCodeLibrary(clock)
+
+        // Act
+        const output = await testFormatter({
+          sources,
+          supportCodeLibrary,
+          type: 'junit',
+        })
+
+        // Assert
+        expect(output).xml.to.deep.equal(
+          '<?xml version="1.0"?>\n' +
+            '<testsuites>\n' +
+            '  <testsuite failures="0" name="my feature" time="0.001" tests="1">\n' +
+            '    <testcase name="my scenario" time="0.001" classname="my-feature;my-scenario">\n' +
+            '      <system-out>🟩 Given a passing step</system-out>\n' +
+            '    </testcase>\n' +
+            '  </testsuite>\n' +
+            '</testsuites>'
+        )
+      })
+    })
+
+    describe('failed', () => {
+      it('includes the error message', async () => {
+        // Arrange
+        const sources = [
+          {
+            data: [
+              'Feature: my feature',
+              '  Scenario: my scenario',
+              '    Given a failing step',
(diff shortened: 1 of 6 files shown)


davidjgoss:
Thanks for raising this @ricalbuquerque, we do indeed need a JUnit formatter and this looks like a good implementation at first glance.

There's a broader effort in https://github.com/cucumber/common/issues/871 to have a JUnit formatter for all Cucumber implementations that behave the same (share a test suite etc), so I'm going to bring this up there and see if it could be a vehicle to get that piece moving a bit in the right direction, and I'll probably come back with some details review comments soon.

ricalbuquerque:
Thank you @davidjgoss. I don't mean to interfere with any plans for a new formatter. I did try a few formatters (some of them seem to be a bit old) and none would do what I was expecting to (treating scenarios as testcases) so this was my motivation.
I did see that issue on the common repo but since it seemed to be a bit blocked I decided to implement a version of it in javascript since I don't have that much experience in Go.
Feel free to suggest any improvements you would like to see!

davidjgoss:
@ricalbuquerque not an interference at all, I think this could help move it forward. I don't think there will be a single implementation in Go, that model didn't work great in practise for other components. More like, a shared set of tests and then native implementations in JavaScript, Java, Ruby etc that all pass those same tests. So this is definitely in line with that.

mpkorstanje (commented):
Looking at the test cases, I feel there may be a lot of fidelity missing. I can't provide concrete examples right now but contrast with the Cucumber JVM test cases:

https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/test/java/io/cucumber/core/plugin/JUnitFormatterTest.java

And so it might be good to compare against the XSD.

https://github.com/cucumber/cucumber-jvm/blob/main/cucumber-core/src/test/resources/io/cucumber/core/plugin/surefire-test-report-3.0.xsd

src/formatter/junit_formatter_spec.ts
```diff
@@ -0,0 +1,245 @@
...
+import { expect, use } from 'chai'
+import chaiXml from 'chai-xml'
+import FakeTimers, { InstalledClock } from '@sinonjs/fake-timers'
+import timeMethods from '../time'
+import { testFormatter } from '../../test/formatter_helpers'
+import { getJsonFormatterSupportCodeLibrary } from '../../test/fixtures/json_formatter_steps'
+
+use(chaiXml)
+
+describe('JunitFormatter', () => {
```
davidjgoss, line 11: Good testing here, there are some cases you'll want to cover as well though:

- pending steps
- skipped test cases
- scenario outline with multiple examples

davidjgoss (changes requested):
Following on from previous comments, just calling out things that are different from the style seen in cucumber-jvm which I think we should align with:

- All steps from the scenario with their statuses are printed with a consistent line width in the `<system-out>` tag, see https://github.com/cucumber/cucumber-jvm/blob/ee6b693184b463c023a265fe98fa9ab5ab2ce819/cucumber-core/src/test/java/io/cucumber/core/plugin/JUnitFormatterTest.java#L69
- There is a single `<testsuite>` (not one for each feature), and each `<testcase>` uses the feature name as `classname` and scenario name as `name` respectively
- Failures have the `type` and `message` as attrs on the `<failure>` tag and the stack in the body of that tag - I don't think we can do this with cucumber-js because the message and stack are not separated, so we'll have to omit `message` attr

I'm happy to jump in and help out with any of this if you'd like, just let me know. Thanks again for putting this together.

tim-yao:
We need this quickly as CircleCI will stop support for JSON format in 20 days.
```
Starting November 20th, all projects using the Cucumber test runner must store test results using the junit format.
```
https://discuss.circleci.com/t/cucumber-test-results-format-deprecation/45776

davidjgoss:
@ricalbuquerque are you happy for me to make some of the changes discussed above so we can get this in?

src/formatter/junit_formatter_spec.ts
```diff
@@ -0,0 +1,245 @@
...
+          supportCodeLibrary,
+          type: 'junit',
+        })
+
+        // Assert
+        expect(output).xml.to.deep.equal(
+          '<?xml version="1.0"?>\n' +
+            '<testsuites>\n' +
+            '  <testsuite failures="0" name="my feature" time="0.001" tests="1">\n' +
+            '    <testcase name="my scenario" time="0.001" classname="my-feature;my-scenario">\n' +
```
mpkorstanje, line 67: Why this concatenation of scenario and feature name in the `classname` field?

src/formatter/junit_formatter_spec.ts
```diff
@@ -0,0 +1,245 @@
...
+            '      <system-out>🟩 Given a passing step</system-out>\n' +
```
mpkorstanje, line 68: I would expect a CDATA element here. 

For examples

```feature
Feature: XML <element />

  Scenario: XML <element />
    Given XML <element />
    And Some <![CDATA[ Nested CDATA elements ]]>
```

mpkorstanje (changes requested):
Looks good! Spotted a few complicated problems.

davidjgoss:
@mpkorstanje I've made some updates to bring this pretty much in line with what the cucumber-jvm one does, and added more tests. Would you mind taking another look before I merge?

davidjgoss (approved):
Huge thanks @ricalbuquerque for contributing this!

aslakhellesoy:
Hi @ricalbuquerque,

Aslak Hellesøy
Creator of Cucumber

mpkorstanje:
> @mpkorstanje I've made some updates to bring this pretty much in line with what the cucumber-jvm one does, and added more tests. Would you mind taking another look before I merge?

Looks good!

PR #2134, opened 2022-09-06
Enchanced features filter

lamartire:
### 🤔 What's changed?

The PR introduces runtime filters to enhance specs filtering not only be `path` or `scenario`, but by any `Pickle` message field:

```js
module.exports = {
  default: {
    include: (pickle) => pickle.tags.includes('foo'),
    exclude: (pickle) => pickle.id === '...',
  }
}
```

With the change we can use some good Allure features such as re-run with for `cucumber-js`.

I'm not sure about the place where I added the fields. I decided to place them into `IRunOptions.runtime` because we use them in the runtime.

### 📋 Checklist:

----

### DIFF
--- a/src/api/convert_configuration.ts
+++ b/src/api/convert_configuration.ts
@@ -31,6 +31,8 @@ export async function convertConfiguration(
       retryTagFilter: flatConfiguration.retryTagFilter,
       strict: flatConfiguration.strict,
       worldParameters: flatConfiguration.worldParameters,
+      include: flatConfiguration.include,
+      exclude: flatConfiguration.exclude,
     },
     formats: convertFormats(flatConfiguration, env),
   }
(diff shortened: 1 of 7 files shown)


davidjgoss:
Thanks for raising @lamartire 

We've been [working on a plugins mechanism](https://github.com/cucumber/cucumber-js/discussions/2091) recently and the use case of "filter/order pickles in a custom way" is one that we identified would be a good fit to enable via plugins.

I just want to understand your use case a bit better though - from some of the things you've said, it seems like you might be running Cucumber programatically as part of a bigger framework, is that the case?

lamartire:
> Thanks for raising @lamartire
> 
> We've been [working on a plugins mechanism](https://github.com/cucumber/cucumber-js/discussions/2091) recently and the use case of "filter/order pickles in a custom way" is one that we identified would be a good fit to enable via plugins.
> 
> I just want to understand your use case a bit better though - from some of the things you've said, it seems like you might be running Cucumber programatically as part of a bigger framework, is that the case?

We want to create seamless integration for `allure` and `cucumber-js` and one of the most important features is filtering specs by unique predicate (much more specific, than path, or name) when we need to perform re-run.

Plugins could be the solution, as far as I see. Could you please share some details about the API to estimate a sense of the PR? :)

PR #2139, opened 2022-09-12
Change the reported url and method

harcop:
### 🤔 What's changed?

Fix the reported url and method

Fix the misleading error message. If the second http request fails, the error message refers to the first request, this change fixes this so that the second request is used instead.

### 📋 Checklist:

----


src/publish/http_stream.ts
```diff
@@ -123,9 +127,14 @@ export default class HttpStream extends Transform {
...
 
     const allHeaders = { ...headers, ...additionalHttpHeaders }
+    // eslint-disable-next-line no-console
+    console.log('HEADERS:', allHeaders)
+    // eslint-disable-next-line no-console
+    console.log('URL:', url)
     const req = httpx.request(url, {
       method,
       headers: allHeaders,
+      timeout: 10000,
```
davidjgoss, line 137: Did adding this solve a problem for you?

davidjgoss (changes requested):
Thanks for getting this going @harcop!

Would you be able to:
- Remove the `.tool-versions` file
- Revert changes to the `package-lock.json` (looks like you have an older version of npm there)
- Remove the `console.log` statements

PR #2227, opened 2023-01-25
fix(junit-formatter): add file name on testcase tag

zcmgyu:
### 🤔 What's changed?

Added `file` attribute on each `testcase` tag of XML-based JUnit report

In CircleCI, in order to optimize Cucumber tests across a set of parallel executors, CircleCI requires test results to be uploaded as JUnit XML format reports with executed timing data. Based on needed well-formed JUnit XML reports, CircleCI parses and uses data for the following test splitting based on timing data taken to execute tests of a specific `test name`, `class name`, or `file name`. [CircleCI required JUnit XML report format as below:](https://circleci.com/docs/use-the-circleci-cli-to-split-tests/#junit-xml-reports)

- The `file` attribute, either on the `<testsuite>` or `<testcase>` tag
- The `time` attribute, on the `<testcase>` tag

``` XML
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Mocha Tests" tests="3" failures="1">
  <testsuite tests="3">
    <testcase classname="foo1" name="ASuccessfulTest" time="10" file="src/__tests__/App.test.js" />
    <testcase classname="foo2" name="AnotherSuccessfulTest" time="5" file="src/__tests__/App.test.js" />
    <testcase classname="foo3" name="AFailingTest" time="1.1050" file="src/__tests__/App.test.js">
        <failure type="NotEnoughFoo"> details about failure </failure>
    </testcase>
  </testsuite>
</testsuites>
```

The `time` attribute is already on the `<testcase>` tag of the XML-based JUnit report, this PR just adds the `file` attribute on the `<testcase>` tag

Feel free to suggest any change or possible improvement.

### 📋 Checklist:

### DIFF
--- a/src/formatter/junit_formatter.ts
+++ b/src/formatter/junit_formatter.ts
@@ -34,6 +34,7 @@ interface IJUnitTestCase {
   result: IJUnitTestCaseResult
   systemOutput: string
   steps: IJUnitTestStep[]
+  file: string
 }
 
 interface IJUnitTestCaseResult {
@@ -221,6 +222,7 @@ export default class JunitFormatter extends Formatter {
           result: this.getTestCaseResult(steps),
           systemOutput: this.formatTestSteps(steps),
           steps,
+          file: pickle.uri,
         }
       }
     )
@@ -257,6 +259,7 @@ export default class JunitFormatter extends Formatter {
         classname: test.classname,
         name: test.name,
         time: test.time,
+        file: test.file,
       })
       if (test.result.status === TestStepResultStatus.SKIPPED) {
         xmlTestCase.ele('skipped')
--- a/src/formatter/junit_formatter_spec.ts
+++ b/src/formatter/junit_formatter_spec.ts
@@ -104,7 +104,7 @@ describe('JunitFormatter', () => {
         expect(output).xml.to.deep.equal(
           '<?xml version="1.0"?>\n' +
             '<testsuite failures="0" skipped="0" name="cucumber-js" time="0.001" tests="1">\n' +
-            '  <testcase classname="my feature" name="my scenario" time="0.001">\n' +
+            '  <testcase classname="my feature" name="my scenario" time="0.001" file="a.feature">\n' +
             '    <system-out><![CDATA[Given a passing step......................................................passed]]></system-out>\n' +
             '  </testcase>\n' +
             '</testsuite>'
@@ -139,7 +139,7 @@ describe('JunitFormatter', () => {
         expect(output).xml.to.deep.equal(
           '<?xml version="1.0"?>\n' +
             '<testsuite failures="1" skipped="0" name="cucumber-js" time="0.001" tests="1">\n' +
-            '  <testcase classname="my feature" name="my scenario" time="0.001">\n' +
+            '  <testcase classname="my feature" name="my scenario" time="0.001" file="a.feature">\n' +
             '    <failure type="FAILED" message="A hook or step failed"><![CDATA[error]]></failure>\n' +
             '    <system-out><![CDATA[Given a failing step......................................................failed]]></system-out>\n' +
             '  </testcase>\n' +
@@ -176,7 +176,7 @@ describe('JunitFormatter', () => {
         expect(output).xml.to.deep.equal(
           '<?xml version="1.0"?>\n' +
             '<testsuite failures="0" skipped="0" name="cucumber-js" time="0.001" tests="1">\n' +
-            '  <testcase classname="my feature" name="my scenario" time="0.001">\n' +
+            '  <testcase classname="my feature" name="my scenario" time="0.001" file="a.feature">\n' +
             '    <system-out><![CDATA[Given a flaky step........................................................passed]]></system-out>\n' +
             '  </testcase>\n' +
             '</testsuite>'
@@ -210,7 +210,7 @@ describe('JunitFormatter', () => {
         expect(output).xml.to.deep.equal(
           '<?xml version="1.0"?>\n' +
             '<testsuite failures="1" skipped="0" name="cucumber-js" time="0.001" tests="1">\n' +
-            '  <testcase classname="my feature" name="my scenario" time="0.001">\n' +
+            '  <testcase classname="my feature" name="my scenario" time="0.001" file="a.feature">\n' +
             '    <failure type="PENDING" message="A step in the test case is not yet implemented"/>\n' +
             '    <system-out><![CDATA[Given a pending step.....................................................pending]]></system-out>\n' +
             '  </testcase>\n' +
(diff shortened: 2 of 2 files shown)


davidjgoss:
Thanks for raising this @zcmgyu.

I'm in two minds because this seems a useful capability to hook up with in CircleCI, but we don't have the `file` attr in our [canonical JUnit formatter](https://github.com/cucumber/cucumber-junit-xml-formatter) from the JVM implementation, and it's not a standard from the (admittedly not particularly well standardised) schema. @mpkorstanje what do you think?

zcmgyu:
@davidjgoss Thank you for considering. It would be great if this functionality could support when running Cucumber by feature files on the CircleCI environment. 🔥

mpkorstanje:
Following the principle of robustness whe should be strict in what we emit. So I would recommend against deviations from the schema.

But giving the Circle CI documentation a cursory read I also don't think we need the changes proposed in this particular PR to support Circle CI test splitting.

Based on the full example provided at the end of the page suggests the following pattern:

```
circleci tests glob "test/**/*.rb" | circleci tests split > /tmp/tests-to-run
bundle exec rspec $(cat /tmp/tests-to-run)
```

Ideally we follow this pattern by allowing Cucumber to list it's scenarios and/or features as they appear in the report.

So something like

```
cucumber-js --list scenarios  circleci tests split --time > /tmp/tests-to-run
cucumber-js $(cat /tmp/tests-to-run)
```

zcmgyu:
Hi! @mpkorstanje 
Thank you for your suggestion. 
Do you think about the Scenario Outline with multiple examples? Suppose that I could list up the Scenario Outline name but in the XML output, the test case will be listed with the index. CircleCI could hook up with the first example, but from the second one, I think CircleCI will have no idea what is this scenario.

Below is the output of Scenario Outline `my templated scenario` executed.

``` xml
<?xml version="1.0"?>
<testsuite failures="1" skipped="0" name="cucumber-js" time="0.002" tests="2">
	<testcase classname="my feature" name="my templated scenario" time="0.001">
		<system-out>
			<![CDATA[Given a passing step......................................................passed]]>
		</system-out>
	</testcase>
	<testcase classname="my feature" name="my templated scenario [1]" time="0.001">
		<failure type="FAILED" message="A hook or step failed">
			<![CDATA[error]]>
		</failure>
		<system-out>
			<![CDATA[Given a failing step......................................................failed]]>
		</system-out>
	</testcase>
</testsuite>
```

mpkorstanje:
I do think it is reasonable to expect Cucumber to render the scenario names with sufficient detail in the XML report so that they can be selected for execution again

For example the canonical XML formatter renders the following feature 

```feature
Feature: Examples Tables

  Scenario Outline: Eating cucumbers
    Given there are <start> cucumbers
    When I eat <eat> cucumbers
    Then I should have <left> cucumbers

    Examples: These are passing
      | start | eat | left |
      |    12 |   5 |    7 |
      |    20 |   5 |   15 |

    Examples: These are failing
      | start | eat | left |
      |    12 |  20 |    0 |
      |     0 |   1 |    0 |
```

In the XML as:
```xml
<testcase classname="Examples Tables" name="Eating cucumbers - These are passing - Example #1.1" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are passing - Example #1.2" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are failing - Example #2.1" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are failing - Example #2.2" />

```

I don't believe we can currently do this round trip in Cucumber JS, but that would be reasonable to implement in one or more PRs.

I also don't know the exact details of the circle CI `circleci tests split --time` command, I couldn't find documentation for it either. So some experimentation may be need to see exactly how it processes it's inputs. It may be that the inputs are not processed in a reasonable way.

mpkorstanje:
And I do realize that this a bigger ask then the current suggestion, I'm sorry about that. In the long run it causes less grief for us to do things the right way.

zcmgyu:
After a week of researching the CircleCI documentation, this is my awareness of CircleCI split API.

Suppose that I have a list of scenario names in the `scenarioList.txt` file.
```
Eating cucumbers
Drink cucumber juice
Cook cucumber
```
And run it in 2 parallel executors.

`cat scenarioList.txt | circleci tests split --split-by=timings`

Based on the timing data from the JUnit XML report of the previous job run, CircleCI adjusts which test case should be in which executor. As you can see, there is no scenario named `Eating cucumbers - These are passing - Example #1.1` in `scenarioList.txt`, then no timing data is found, the tests will then be split alphabetically by test name.

``` xml
<testcase classname="Examples Tables" name="Eating cucumbers - These are passing - Example #1.1" time="30" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are passing - Example #1.2" time="60" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are failing - Example #2.1" time="20" />
<testcase classname="Examples Tables" name="Eating cucumbers - These are failing - Example #2.2" time="50" />
...
```

How about if we could list the case as a JUnit XML report in the `scenarioList.txt` file?

```
Eating cucumbers - These are passing - Example #1.1
Eating cucumbers - These are passing - Example #1.2
...
```
Now, CircleCI already knows `Eating cucumbers - These are passing - Example #1.1` test case took 30 seconds to finish.
I have no idea how to filter which scenario outline's example is run in CucumberJS.
https://github.com/cucumber/cucumber-js/blob/main/docs/filtering.md

I did a trick by slicing every example to multiple feature files, then I could filter which example is run by file name. But I missed the file name in the JUnit XML report.

One more odd scenario I'm always wondering about, some examples of the same Scenario Outline could not run in parallel.  I could not list it as an execution unit, it got a conflict while running in parallel.

```
Eating cucumbers - These are passing - Example #1.1
Eating cucumbers - These are passing - Example #1.2
...
```

I know it is tough to make sense of all common scenarios. Firstly, I just patch this case in my project.

Let me know if you want some helps. 

Thanks again 💯 👍🏻

mpkorstanje:
> One more odd scenario I'm always wondering about, some examples of the same Scenario Outline could not run in parallel. I could not list it as an execution unit, it got a conflict while running in parallel.

Parallel execution of tests is a relatively complex implementation wise. So Cucumber was build with the assumption is that each example and scenario would be independent.

And in the context of Circle CI, we have to flatten Cucumbers hierarchy of scenarios and examples into a `classname method-name` format for the XML report so it makes sense to hold onto this assumption.

Fundamentally this problem isn't unsolvable, for example in the Java ecosystem with JUnit 5 a generic platform was developed for the discovery, selection and execution of tests. Within this framework tests can indicate which resources they need. A test splitting/distribution tool can use this information to make safe groupings. I'm not aware of such a framework in the JavaScript ecosystem. I also think it extremely unlikely we'll be the ones to develop it.

zcmgyu:
> And in the context of Circle CI, we have to flatten Cucumbers hierarchy of scenarios and examples into a classname method-name format for the XML report so it makes sense to hold onto this assumption.

It sounds great.

PR #2229, opened 2023-02-04
feat: include exceptions in test step result messages

davidjgoss:
### 🤔 What's changed?

- Include `exception` in the result where there was an error in running a step.
- Use that in the JUnit formatter for the `<failure/>` tag

Confirm with latest CCK behaviour, get better parity with JVM's JUnit formatter

### 📋 Checklist:

----


compatibility/features/cdata/cdata.ts
```diff
@@ -0,0 +1,9 @@
+import assert from 'assert'
+import { Given } from '../../../src'
+
+Given(
+  'I have {int} <![CDATA[cukes]]> in my belly',
```
mpkorstanje, line 5: I can't find the matching expectation for this one.

davidjgoss: This runs against the corresponding CCK feature https://github.com/cucumber/compatibility-kit/tree/main/devkit/samples/cdata

mpkorstanje: But the CCK doesn't contain XML to assert against. I would expect this outcome:

https://github.com/cucumber/cucumber-junit-xml-formatter/blob/main/testdata/cdata.feature.xml

davidjgoss: Ah okay, so that's a test case we can add to the JUnit formatter for now, and then once it moves to the polyglot repo it will have that test anyway.

mpkorstanje: 👍

davidjgoss: Pushed.

mpkorstanje (approved):

Rubea1435:
Can you please explain what's changed in this pull request? I use Cucumber.js since 8.0.0 version and there always have been error messages in step results with error trace.

davidjgoss:
@Rubea1435 you're right in that the information was there before. It's just now included in a more structured format that makes it more useful to certain reporting tools.

Issue #2239, opened 2023-02-12
make json formatter reusable outside cucumber-js


davidjgoss:
### 🤔 What's the problem you're trying to solve?

Other tools in the JavaScript ecosystem would like to be able to generate the legacy JSON format from messages, without asking their users to deal with the Go-based json-formatter - https://github.com/cucumber/json-formatter/issues/13#issuecomment-1242690941

### ✨ What's your proposed solution?

- Refactor the JSON formatter to just deal with messages, like e.g. the [HTML one does](https://github.com/cucumber/html-formatter), and the [JUnit one eventually will](https://github.com/cucumber/cucumber-junit-xml-formatter). Do this in a way where it's reusable outside the context of cucumber-js i.e. just requiring `messages` and maybe `query`.
- Probably publish it as a separate npm package
- _Maybe_ publish it in the [mooted new plugin/formatter style](https://github.com/cucumber/cucumber-js/discussions/2091#discussioncomment-3211249) - I'd like for formatters to be usable by any messages-emitting tool, not just cucumber-js.

### ⛏ Have you considered any alternatives or workarounds?

Doing nothing is another choice. But this helps the Cypress implementation and also helps move forward the new plugin/formatter system with something concrete.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

PR #2240, opened 2023-02-12
deprecation updates ahead of 9.0.0

davidjgoss:
### 🤔 What's changed?

- Some rewording and fixing around how we describe deprecations
- Undeprecate `defineStep` - at the time the direction seemed to be to eventually make [strict keywords](https://github.com/cucumber/common/issues/768) the only behaviour. However there is [not a consensus across platforms ](https://github.com/cucumber/cucumber-jvm/issues/2557#issuecomment-1137867515) for doing that, so on that basis it should only be an option. The `PickleStep` model [has an 'Unknown' value](https://github.com/cucumber/messages/blob/main/jsonschema/Pickle.json#L50) so we can use that for `defineStep` and just throw if you try to use it with strict keywords. Finally, a user raised [a separate concern](https://github.com/cucumber/cucumber-js/issues/2043#issuecomment-1267136955) in defense of `defineStep` from a linguistics perspective. So overall I think there's no longer a case to remove this.
- ~~Deprecate the export of `JsonFormatter` class - currently most of the built-in formatter classes are exported like this but I think they should mostly be considered implementation details. When we tackle https://github.com/cucumber/cucumber-js/issues/2239 we'll rewrite this formatter, probably in the new "plugin"-style format we've discussed, so we should get this off the entry point as soon as we can.~~
    - Whilst it would be good to move early on this, we can't reasonably deprecate something without pointing to the new thing that people should migrate to, which we don't have yet, so omitting this for now.

### 📋 Checklist:

----

### DIFF
--- a/docs/deprecations.md
+++ b/docs/deprecations.md
@@ -22,14 +22,6 @@ The `Cli` class is used internally to represent an instance of the command-line 
 
 To adapt, pivot to the `runCucumber` function from the [JavaScript API](./javascript_api.md), or raise an issue if you feel your use case isn't catered for.
 
-### `defineStep`
-
-Deprecated in `8.3.0`, will be removed in `10.0.0` or later.
-
-`defineStep` is a way to generically define a step without tying it to any of the `Given`, `When` or `Then` keywords. This leads to ambiguity in the business language and is considered an antipattern; steps should be clearly applicable to one of those keywords, being a context, action or outcome respectively.
-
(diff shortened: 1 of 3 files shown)


mpkorstanje:
From a process point of view I have only one remark:

> except maybe SummaryFormatter which is the one people would reasonably extend when building their own.

If this class is open for extension it should be documented as such. And be tested against the specific use cases it aims to support.

PR #2244, opened 2023-02-15
2243 clarify transpile.md

michael-lloyd-morris:
### 🤔 What's changed?

Clarified the differences between transpiling Cucumber in CommonJS and ESM

Spent 4 hours slogging through this.

Is this rewrite clear to others?

### 📋 Checklist:

----

### DIFF
--- a/docs/transpiling.md
+++ b/docs/transpiling.md
@@ -1,6 +1,6 @@
 # Transpiling
 
-Step definitions and support files can be written in a syntax or language that compiles to JavaScript, and just-in-time compiled when you run Cucumber. The output of the transpiler must match the module format expected by node, and you must use the correct Cucumber directive to import the code, "import" for esm and "require" for CommonJS.
+Step definitions and support files can be written in a syntax or language that compiles to JavaScript, and just-in-time compiled when you run Cucumber. The output of the transpiler must match the module format expected by Node.js, and you must use the correct Cucumber directive to import the code, "import" for ESM and "require" for CommonJS.
 
 For example, you might want to use [Babel](https://babeljs.io/):
 
@@ -19,7 +19,7 @@ Your `tsconfig.json` should have these `compilerOptions`:
 
 ```
 
-Typescript's output must match Cucumber's expected import, and this is controlled by the "module" option in the `tsconfig.json`.  The default is "CommonJS", but if your project is setup to be an ESM project then Typescript will need to output some generation of ES code to Cucumber. If you aren't sure use "ESNext".
+Typescript's output must match Cucumber's expected input, and this is controlled by the "module" option in the `tsconfig.json`.  The default is "CommonJS", but if your project is setup to be an ESM project then Typescript will need to output some generation of ES code to Cucumber. If you aren't sure use "ESNext".
 
 You'll also need to specify where your support code is, since `.ts` files won't be picked up by default.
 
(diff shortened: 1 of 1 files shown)


Jeff-Tian:
Thanks, @michael-lloyd-morris! Great work, it'll benefit many others unfamiliar with nodejs and esm etc.

Jeff-Tian (approved):
Looks good to me! Very helpful messages indeeded.

docs/transpiling.md
```diff
@@ -31,10 +33,27 @@ If you are using [ts-node](https://github.com/TypeStrong/ts-node):
...
+
+To use ESM modules set the `"type": "module"` in the package.json file.
 
-```shell
-$ NODE_OPTIONS="--loader ts-node/esm" cucumber-js --import 'step-definitions/**/*.ts'
+Your cucumber.js file will need to be an ESM module and contain the following at a minimum:
+
+```javascript
+export default {
+  requireModule: ['ts-node/register'],
```
davidjgoss, line 54: This doesn’t seem right? require-module doesn’t affect import

docs/transpiling.md
```diff
@@ -31,10 +33,27 @@ If you are using [ts-node](https://github.com/TypeStrong/ts-node):
...
+    "test": "cross-env NODE_OPTIONS=\"--loader ts-node/esm\" cucumber-js"
+  }
+}
+```
+
+To use ESM modules set the `"type": "module"` in the package.json file.
 
-```shell
-$ NODE_OPTIONS="--loader ts-node/esm" cucumber-js --import 'step-definitions/**/*.ts'
+Your cucumber.js file will need to be an ESM module and contain the following at a minimum:
```
davidjgoss, line 50: The config file can still be JSON or YAML - I guess we want to say that if JavaScript then it should be the appropriate format.

docs/transpiling.md
davidjgoss: I think this can just be removed.

docs/transpiling.md
```diff
@@ -31,10 +33,27 @@ If you are using [ts-node](https://github.com/TypeStrong/ts-node):
...
 #### ESM
 
-For ESM projects, you can use `ts-node`'s ESM loader and then `import` your TypeScript files:
+For ESM projects, you can use `ts-node`'s ESM loader and then import your TypeScript files. Doing this will require setting an environment variable. the cleanest way to do this is to include the [cross-env](https://www.npmjs.com/package/cross-env) package with `npm i -D cross-env`
+
+Then change your npm test script invocation in the package.json file to:
+
+```json
+{
+  "scripts: {
```
davidjgoss, line 42: Missing a double quote on the key here

michael-lloyd-morris: Ah.  Sorry.  Also spotted an uncapitalized T at the start of a sentence so I sent the whole thing through a spell/grammar checker to be sure. Should be good now.

docs/transpiling.md
michael-lloyd-morris: Lines removed.

docs/transpiling.md
michael-lloyd-morris: It needs to be reworded heavily and I've done so. If you read that header as applying to all ESM projects others will too. That sub header is level 4 and it was meant to cover "with TS-Node to load ES modules"

michael-lloyd-morris:
These examples are fairly complex and they also raise a lot of Google and Bing search results. I have a solution, but it will affect more than this - setup a group of Cucumber-JS scenario repositories that address the most common applications of the system one at a time.

michael-lloyd-morris:
@davidjgoss Please let me know what else do I need to do for this.

docs/transpiling.md
```diff
@@ -1,52 +1,53 @@
...
 "allowSyntheticDefaultImports": true,
 "resolveJsonModule": true,
+
 ```
 
-Other than that, a pretty standard TypeScript setup should work as expected.
+Typescript's output must match Cucumber's expected import, and this is controlled by the "module" option in the `tsconfig.json`.  The default is "CommonJS", but if your project is setup to be an ESM project then Typescript will need to output some generation of ES code to Cucumber. If you aren't sure use "ESNext".
 
-You'll also need to specify where your support code is, since `.ts` files won't be picked up by default.
+Cucumber doesn't load `*.ts` files by default. The glob pattern `step-definitions/**/*.ts` will load them from the default step definition location.
```
davidjgoss, line 24: I don't think this line is an improvement. `step-definitions` isn't any of our default paths.

docs/transpiling.md
```diff
@@ -1,52 +1,53 @@
...
 ### With ts-node
 
-If you are using [ts-node](https://github.com/TypeStrong/ts-node):
+[TS-Node](https://github.com/TypeStrong/ts-node): is the one of the most popular ways to load TypeScript files. 
 
 - In a configuration file `{ requireModule: ['ts-node/register'], require: ['step-definitions/**/*.ts'] }`
 - On the CLI `$ cucumber-js --require-module ts-node/register --require 'step-definitions/**/*.ts'`
 
-#### ESM
+If you are using ts-node in a CommonJS project then this configuration will work, but if you have an ESM project there are additional steps.
```
davidjgoss, line 35: We should make clear that both `requireModule` and `require` are not suitable for ESM and that the `--loader` flag and `import` option are their respective equivalents.

docs/transpiling.md
```diff
@@ -1,52 +1,53 @@
...
-For ESM projects, you can use `ts-node`'s ESM loader and then `import` your TypeScript files:
+* Set TypeScript to export to an ES format such as "ESNext" using the `ts-config.json` file. 
+* Then use `ts-node`'s ESM loader to import your TypeScript. 
 
-```shell
-$ NODE_OPTIONS="--loader ts-node/esm" cucumber-js --import 'step-definitions/**/*.ts'
-```
-
-Don't forget to set your `tsconfig.json` to emit JavaScript with `import` and `export` statements:
+That last step requires setting an environment variable. The cleanest way to do this is to include the [cross-env](https://www.npmjs.com/package/cross-env) package with `npm i -D cross-env`. With that package installed make the following change to your npm test script invocation in the package.json file:
```
davidjgoss, line 40: I don't think we should be prescriptive of using this particular library/approach. By all means mention it as an option but examples should be as vanilla as possible really.

michael-lloyd-morris: Done.

docs/transpiling.md
michael-lloyd-morris: Done.

davidjgoss (approved):

PR #2245, opened 2023-02-18
stop using deprecated api in feature tests

charlierudolph:
Currently our CI output includes this warning:

```
(node:2090) DeprecationWarning: `Cli` is deprecated, use `runCucumber` instead; see https://github.com/cucumber/cucumber-js/blob/main/docs/deprecations.md
(Use `node --trace-deprecation ...` to show where the warning was created)
```

Updated cucumber-js feature tests to stop using our own deprecated API


features/support/world.ts
```diff
@@ -83,16 +85,20 @@ export class World {
...
+          provided: argvConfiguration,
+        },
+        environment
+      )
       let error: any
       try {
-        const { success } = await cli.run()
+        const { success } = await runCucumber(runConfiguration, environment)
         if (!success) {
           error = new Error('CLI exited with non-zero')
```
davidjgoss, line 103: Should this now say something like "runCucumber threw an error"?

features/support/world.ts
```diff
@@ -83,16 +85,20 @@ export class World {
...
       const stderr = new PassThrough()
-      const cli = new Cli({
-        argv: args,
-        cwd,
-        stdout,
-        stderr,
-        env,
-      })
+      const environment: IRunEnvironment = { cwd, stdout, stderr, env }
+      const { options, configuration: argvConfiguration } =
```
davidjgoss, line 89: Looking at the 2 failing scenarios, I think this call to `parse` and the one below to `loadConfiguration` should also move inside of the try block.

davidjgoss (commented):
Looks good, couple of suggestions.

davidjgoss:
When first implementing `runCucumber` I looked at this and half-considered just making `@spawn` the only behaviour i.e. so they all run via the executable, but that would probably slow it down a lot.

features/support/world.ts
charlierudolph: Updated

features/support/world.ts
charlierudolph: Moved inside try block

davidjgoss (approved):

PR #2246, opened 2023-02-19
run prettier directly

charlierudolph:
When I originally setup prettier, we ran it as part of eslint, however prettier does not recommend this: https://prettier.io/docs/en/integrating-with-linters.html

Not sure if I missed that guidance originally or didn't look for it. 

Anyway, by running it standalone, we can have it standardize formatting in more things. Suggest reviewing this by commit and first commit is my manual changes and then a couple of rounds of running `npm run lint-format-autofix` (not sure exactly why it took 2 rounds).


.prettierignore
```diff
@@ -0,0 +1,11 @@
...
+docs
+lib
+node_modules
+reports
+tmp
+
+CHANGELOG.md
+CONTRIBUTING.md
+README.md
+UPGRADING.md
```
charlierudolph, line 11: Think we do probably want to extend to running prettier on these md files and most of docs. But thinking can enable that in a follow up

davidjgoss: Makes sense. We'd need to watch out for `CHANGELOG.md` as that gets written by a robot.

davidjgoss (approved):

PR #2251, opened 2023-03-01
remove my work e-mail from package.json

GeekyDeaks:
Hi, sorry but I have been asked by my company to remove my work e-mail address from all public repos and this is one of them.

This was my mistake as I was accidentally using the wrong account when contributing to the project.

Cheers,
Scott

### DIFF
--- a/package.json
+++ b/package.json
@@ -142,6 +142,7 @@
     "RolandArgos <roland.ormrod@argos.co.uk>",
     "Ronald Chen (https://github.com/Pyrolistical)",
     "Sam Saccone <sam@samx.it>",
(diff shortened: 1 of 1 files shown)


davidjgoss:
No problem @GeekyDeaks. We can still include you with your GitHub profile URL instead of email address if that would be okay? (I have started doing this for all new contributors, for exactly this kind of reason.)

GeekyDeaks:
Yeah, that works for me.  It's just the e-mail address that is the problem.  Thanks for being so understanding on this

davidjgoss (approved):

PR #2261, opened 2023-03-20
Base64 encode string literals in JsonFormatter

### LINKED ISSUE #2260: Missing content encoding for attachments in JSON reports
### 👓 What did you see?

Consider the following code adding two string attachments:

```js
Given("a step", function () {
  const myString = "foobar"
  this.attach(myString, "text/plain");
  this.attach(Buffer.from(myString).toString("base64"), "base64:text/plain");
});
```

This will create two envelopes (messages) with different `contentEncoding` (`IDENTITY` and `BASE64` respectively). However, this information (content encoding) is lost in the JSON report and the original data cannot be reproduced with certainty.

Below is the JSON report:

```json
[
  {
    "description": "",
    "elements": [
      {
        "description": "",
        "id": "a-feature;an-example",
        "keyword": "Example",
        "line": 2,
        "name": "an example",
        "steps": [
          {
            "arguments": [],
            "keyword": "Given ",
            "line": 3,
            "name": "a step",
            "match": {
              "location": "features/definitions.js:3"
            },
            "result": {
              "status": "passed",
              "duration": 492577
            },
            "embeddings": [
              {
                "data": "foobar",
                "mime_type": "text/plain"
              },
              {
                "data": "Zm9vYmFy",
                "mime_type": "text/plain"
              }
            ]
          }
        ],
        "tags": [],
        "type": "scenario"
      }
    ],
    "id": "a-feature",
    "line": 1,
    "keyword": "Feature",
    "name": "a feature",
    "tags": [],
    "uri": "features/foo.feature"
  }
]
```

I guess for JSON reports to either only ever contain base64 encoded data _or_ for content encoding to be propagated to the report.

* cucumber-js: 9.0.1

1. `git clone https://github.com/badeball/cucumber-reproducible-issues.git`
2. `cd cucumber-reproducible-issues/string-attachment-ambiguity`
3. `npm install`
4. `npx cucumber-js`

badeball:
### 🤔 What's changed?

This fixes #2260.

This ensures that all embeddings in JSON reports are base64-encoded. This is also how [cucumber-json-formatter](https://github.com/cucumber/json-formatter) interprets `IDENTITY` strings.

In the JSON output, it's currently impossible to differentiate between strings that were attached as literals vs. those that were base64-encoded to begin with.

### 📋 Checklist:

----


src/formatter/json_formatter.ts
```diff
@@ -293,7 +293,11 @@ export default class JsonFormatter extends Formatter {
     }
     if (testStepAttachments?.length > 0) {
       data.embeddings = testStepAttachments.map((attachment) => ({
-        data: attachment.body,
+        data:
+          attachment.contentEncoding ===
+          messages.AttachmentContentEncoding.IDENTITY
+            ? Buffer.from(attachment.body).toString('base64')
```
badeball, line 299: Is browser compatibility a requirement for this code? (`Buffer` is node)

davidjgoss: No, we're Node.js only at the moment so `Buffer` is fine.

davidjgoss (approved):

PR #2263, opened 2023-03-20
Issue 2262 ES imports in parallel

michael-lloyd-morris:
### 🤔 What's changed?

Optimization in the support.ts file to gather the dynamic import Promises into an array and then resolve them with a single Promise.all() call.

This should yield a noticeable speedup on Cucumber installs with very large ES step definition libraries. Other Cucumber installs will be unaffected.

- 🐎 Optimization - A change meant to make the code more performant.

If there are other instances in the codebase where an await is inside a loop then this pattern should be applied there as well.

### 📋 Checklist:

### DIFF
--- a/src/api/support.ts
+++ b/src/api/support.ts
@@ -24,10 +24,13 @@ export async function getSupportCodeLibrary({
     requirePaths,
     importPaths,
   })
+  const importPromises:Array<Promise<void>> = []
+
   requireModules.map((module) => require(module))
   requirePaths.map((path) => require(path))
-  for (const path of importPaths) {
-    await importer(pathToFileURL(path))
-  }
+  importPaths.map((path) => importPromises.push(importer(pathToFileURL(path))))
+
+  await Promise.all(importPromises)
+
   return supportCodeLibraryBuilder.finalize()
 }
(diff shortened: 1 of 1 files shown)


michael-lloyd-morris:
There is a behavior change in that the step files resolve simultaneously, but I can't imagine it mattering because Cucumber never gave the end users a way to control the order their step files are resolved in. This might be a concern so I sharded this optimization off to it's own PR for review.

michael-lloyd-morris:
@davidjgoss raised the important point that there may be code out there relying on this sequential load behavior.  I'm not so sure this is the case, but that is the safest approach to this optimization.

That would mean delaying this to the next major release and/or using a cli or config flag to allow opt in.

I'm inclined to follow David's advice and just defer to the next major, especially since require imports are going to remain sequential as that is the nature of the beast. Also, the optimization isn't worth the headache except perhaps for the very largest of test libraries.

michael-lloyd-morris:
Agreed, so flag for version 10 and I'll come back to this when work on that major starts to catch up anything that fell behind?  How do I do that or do you want to do it?

davidjgoss:
Noting that this would, as it stands, break the parallel runtime because that depends on support code being registered in a consistent order across coordinator and workers. We'll need to try and find another way around that in order to go ahead with this change.

PR #2264, opened 2023-03-20
Issue 2105 more detailed error

michael-lloyd-morris:
### 🤔 What's changed?

The error message viewed when importing a CommonJS module or requiring an ES module is now customized.

Making life easier when simple mistakes are made. The error message Node gives when you try to import steps using CommonJS format or require an ES module are a bit cryptic.  The new error message tells the user they are using the wrong directive and to either change the file's format or use the other directive to include the file.

### 📋 Checklist:


davidjgoss (changes requested):
One thing we need to account for here is that, unlike the other way around, it's perfectly valid to `import` a CommonJS module.

The example in the feature file yields an error from Node.js because the file is evaluated as an ES module (due to `.mjs` suffix) but contains `require()` calls so is effectively invalid JavaScript from the engine's perspective. If that file was `.cjs` or even `.js` (assuming no `type="module" on the package) it would work fine.

I _think_ the issue can also show up where you use an ESM loader but it outputs code with `require()` in - this could be easily done if you misconfigure `ts-node` for example.

It's probably worth catching that error still in the `importer`, but I think the messaging will need to be more along the lines of "it looks like some of your support code is using require syntax where only imports are expected" and a link to learn more.

michael-lloyd-morris:
Requested Changes applied.

michael-lloyd-morris:
Is there something else I need to do or where the changes I did insufficient?

src/api/support.ts
```diff
@@ -24,10 +37,13 @@ export async function getSupportCodeLibrary({
     requirePaths,
     importPaths,
   })
-  requireModules.map((module) => require(module))
-  requirePaths.map((path) => require(path))
+
+  requireModules.map((module) => tryRequire(module))
```
davidjgoss, line 41: Can we make the equivalent change for the parallel worker https://github.com/cucumber/cucumber-js/blob/main/src/runtime/parallel/worker.ts#L77-L78 as well?

(At some point it'd be good to extract some code common across these places, but not essential for this PR.)

src/api/support.ts
```diff
@@ -6,6 +6,19 @@ import { pathToFileURL } from 'url'
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const { importer } = require('../importer')
 
+function tryRequire(path: string) {
+  try {
+    return require(path)
+  } catch (error) {
+    if (error.code === 'ERR_REQUIRE_ESM') {
+      throw new Error(`Cucumber expected a CommonJS module at '${path}' but found an ES module.
```
davidjgoss, line 14: It might be worth adding the original error via `cause` option so the user gets the full stack e.g. `new Error('...', { cause: error })` as in some cases it could be down to a file that is imported by `path` rather than that one itself, if that makes sense.

davidjgoss:
Just a couple of things to address but this is looking like a good change. Also see a conflict on the changelog file.

michael-lloyd-morris:
I'll get to these items in the morning and will try to be done by end of day, though I've got to take a chunk of the day to take my Dad to the hospital.

(Until I have a job, volunteer work here is my job - if only to keep my skills sharp and my sanity intact - so I'm treating it as such).

src/api/support.ts
michael-lloyd-morris: Done.  Moved the require logic to its own file in the same directory as the importer and named it try_require.  I see an opportunity for further refactoring, but I agree that should probably be another issue ticket.

src/api/support.ts
davidjgoss: Oh good spot, hadn't realised that was so new. We'll wait until after dropping Node.js 14 support to use that. I'd still like to include the `message` from the error that we catch as part of our message though, so they can see that path if it is nested (the stack doesn't matter so much).

michael-lloyd-morris: Ok. Done.

davidjgoss (approved):

PR #2266, opened 2023-03-22
Cucumber now creates whatever directories are necessary to make a report

### LINKED ISSUE #2159: Reporters could create sub-directory automatically instead of failing

`Cucumber` is unable to create report(s) if a sub-directory structure is specified and the same it's not present; e.g.: `html:reports/cucumber-report.html` — if `${CWD}/reports/` does not exist, the execution fails.

`Cucumber` should be able to create that path, something similar to `mkdir --parents ./reports/`.

Currently I'm using `mkdirp` module to create this path beforehand, but that requires some "orchestration" between `npm` scripts.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

michael-lloyd-morris:
### 🤔 What's changed?
```
  Scenario: Created relative path
    When I run cucumber-js with `-f summary:some/long/path/for/reports/summary.txt`
    Then the file "some/long/path/for/reports/summary.txt" has the text:
      """
      1 scenario (1 passed)
      1 step (1 passed)
      <duration-stat>
      """
```
Previously Cucumber wouldn't created long directory chains as seen in this feature - some outside process had to make sure the directory the report file was to be written into exists.

Fixes #2159 

There was a test in place to see if Cucumber will fail if the directory didn't exist that was removed.  It is possible for this feature to fail if Cucumber lacks permission to create directories - but I feel testing this is a bit of overkill as such permission checking wasn't previously being tested and in any event it's a feature of the OS, not Cucumber.

### 📋 Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
-
+- Reporters create sub-directory automatically instead of failing ([#2266](https://github.com/cucumber/cucumber-js/pull/2266))
 ## [9.0.1] - 2023-03-15
 ### Fixed
 - Ensure feature paths are properly deduplicated ([#2258](https://github.com/cucumber/cucumber-js/pull/2258))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
+### Added
 - Formatters create sub-directory automatically instead of failing ([#2266](https://github.com/cucumber/cucumber-js/pull/2266))
+
 ## [9.0.1] - 2023-03-15
 ### Fixed
 - Ensure feature paths are properly deduplicated ([#2258](https://github.com/cucumber/cucumber-js/pull/2258))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
-- Reporters create sub-directory automatically instead of failing ([#2266](https://github.com/cucumber/cucumber-js/pull/2266))
+- Formatters create sub-directory automatically instead of failing ([#2266](https://github.com/cucumber/cucumber-js/pull/2266))
 ## [9.0.1] - 2023-03-15
 ### Fixed
 - Ensure feature paths are properly deduplicated ([#2258](https://github.com/cucumber/cucumber-js/pull/2258))
--- a/src/api/formatters.ts
+++ b/src/api/formatters.ts
@@ -9,6 +9,7 @@ import fs from 'mz/fs'
 import path from 'path'
 import { IRunOptionsFormats } from './types'
 import { ILogger } from '../logger'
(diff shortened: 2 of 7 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
-
+- Reporters create sub-directory automatically instead of failing ([#2266](https://github.com/cucumber/cucumber-js/pull/2266))
```
davidjgoss, line 11: Can we say "formatters" for consistency

src/api/formatters.ts
```diff
@@ -70,12 +71,26 @@ export async function initializeFormatters({
...
-    formatters.push(await initializeFormatter(stream, target, type))
-  }
+  const streamPromises: Promise<void>[] = []
+
+  Object.entries(configuration.files).forEach(([target, type]) => {
+    streamPromises.push(
+      (async (target, type) => {
+        const absoluteTarget = path.resolve(cwd, target)
+
+        if (!(await fs.exists(absoluteTarget))) {
```
davidjgoss, line 81: `fs.exists` is deprecated, and `mkdirp` is going to be [doing this check anyway](https://github.com/isaacs/node-mkdirp#implementation), so I think it would make more sense to just call `mkdirp` unconditionally.

However I do think we should wrap in a try/catch to be safe, and if something goes wrong do e.g. `logger.warn('Failed to ensure directory for formatter target exists')` and continue.

davidjgoss (commented):
Looks good, couple of requests.

src/api/formatters.ts
michael-lloyd-morris: Requested changes applied.

CHANGELOG.md
michael-lloyd-morris: Change applied.

src/formatter/builder.ts
```diff
@@ -102,13 +102,13 @@ const FormatterBuilder = {
     descriptor: string,
     cwd: string
   ) {
-    let normalised: URL | string = descriptor
+    let normalized: URL | string = descriptor
```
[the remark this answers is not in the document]
davidjgoss: Yeah this is fine, I probably wrote that being a Brit, but if we standardise on anything it should be US.

davidjgoss:
I'd like to merge this but after reverting the last change. It expands the scope of errors we'll tolerate to things like the formatter class failing to load, or an error in its constructor etc, which isn't really about the target dir/file any more.

michael-lloyd-morris:
Well, if I remove the try/catch block then we return to the prior behavior of Cucumber halting with an ENOT error if the file can't be created. Shall I do that?

davidjgoss:
Not removing the try/catch, but reducing its scope. Reverting https://github.com/cucumber/cucumber-js/pull/2266/commits/26212c4c57841d65cf5359ce06e6e63538c8cdf7 would get it back down to just the attempt to ensure the directory i.e. the new behaviour. That feels like the right step for this PR. I'm not sure if failing to create the stream for a formatter should be downgraded to a warning at this point.

michael-lloyd-morris:
I'll change to immediate die so you can see what that looks like. If that isn't good I'll do a full revert back to the point you were comfortable with.

davidjgoss:
Thanks, I think going back to https://github.com/cucumber/cucumber-js/pull/2266/commits/cf5d171757bc05ec0b24360e6f092fa98dd70dfb is the right thing for now

michael-lloyd-morris:
Ok. Done (Well, I used CTRL-Z in Visual Studio code to rewind back to that point instead of git)

davidjgoss (approved):
Thanks for doing this!

PR #2267, opened 2023-03-22
Address deprecation warning in Unit Tests

michael-lloyd-morris:
### 🤔 What's changed?

Fixing a deprecation warning in the unit tests.

```
> mocha 'src/**/*_spec.ts'



  ........(node:38244) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
...........................................................................................................
  ...................................................................................................................
  ................................................................................................

  326 passing (4s)
```

Bugfix - well, a potential bug when the Node team finally follows up on the deprecation.

### 📋 Checklist:

### DIFF
--- a/src/api/test_helpers.ts
+++ b/src/api/test_helpers.ts
@@ -32,6 +32,9 @@ export async function setupEnvironment(): Promise<Partial<IRunEnvironment>> {
 }
 
 export async function teardownEnvironment(environment: IRunEnvironment) {
-  await fs.rmdir(environment.cwd, { recursive: true })
-  environment.stdout.end()
+  return new Promise((resolve) => {
+    fs.rm(environment.cwd, { recursive: true }, resolve)
+  }).then(() => {
+    environment.stdout.end()
+  })
 }


src/api/test_helpers.ts
```diff
@@ -32,6 +32,9 @@ export async function setupEnvironment(): Promise<Partial<IRunEnvironment>> {
 }
 
 export async function teardownEnvironment(environment: IRunEnvironment) {
-  await fs.rmdir(environment.cwd, { recursive: true })
-  environment.stdout.end()
+  return new Promise((resolve) => {
```
michael-lloyd-morris, line 35: Apparently mx/fs doesn't have the new version of fs.rm in its implementation - that or it's incompatible with the old one the deprecation warns about. In any event, to resolve the deprecation I had to resort to the older Promise.then pattern.

PR #2362, opened 2023-12-21
Prevent mutations on world parameters leaking between test cases

### LINKED ISSUE #2356: world parameters are mutable from scenarios, and changes leak to other scenarios
### 👓 What did you see?

For below scenarios, when i try to assign color to world parameter from first scenario as blue, it get retained until the red color from second scenario get assigned. 

Scenario: Will pass
  Given my color is "blue"
  Then my color should not be red

Scenario: Will fail
  Given my color is "red"
  Then my color should not be red  

When world is isolated for each scenario, shouldn't the parameters.color get undefined for second scenario before getting assigned with color red. Currently it is retaining the color blue from previous scenario

@cucumber/cucumber: 10.0.1

davidjgoss:
### 🤔 What's changed?

Clone the world parameters before creating each world instance. This prevents user code that updates the world parameters from leaking those changes across different test cases.

Also, type the `worldParameters` as `JsonObject` in configuration to enforce it should be JSON serialisable.

It would have been nice to somehow ensure immutability even between hooks/steps in the same test case. But realistically, users can have custom world instances, they can do what they want with the `parameters` object they are passed, so there's only so much we can do as the orchestrator.

Fixes #2356 

### 📋 Checklist:

  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 Please see [CONTRIBUTING.md](./CONTRIBUTING.md) on how to contribute to Cucumber.
 
 ## [Unreleased]
+### Fixed
+- Prevent mutations on world parameters leaking between test cases ([#2362](https://github.com/cucumber/cucumber-js/pull/2362))
 
 ## [10.0.1] - 2023-10-20
 ### Fixed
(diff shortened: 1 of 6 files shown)


davidjgoss:
Released in https://github.com/cucumber/cucumber-js/releases/tag/v10.1.0

Issue #2009, opened 2022-04-24
Allow dinamycally defining profiles on esm configs


notaphplover:
### 🤔 What's the problem you're trying to solve?

As developer I would love to dinamycally define profiles on cucumber es module configs.

Actually, the only way to define profiles is exporting the configuration associated to the profile:

**cucumberConfig.js**
```js
/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const profile1 = getConfiguration('profile1');

/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const profile2 = getConfiguration('profile2');

/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const defaultCucumberConfig = getConfiguration();

export default defaultCucumberConfig;

export { profile1, profile2 };
```

If I have several dinamycally generated configs, I can't figure out any way to export them in an esm with the current `cucumber-js` profile configuration.

### ✨ What's your proposed solution?

There's not a single and best way to achieve this imho. Maybe a default export should be mandatory and it would contain a config or an object with profiles, being `default` the default one:

**cucumberConfig.js**
```js
/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const profile1 = getConfiguration('profile1');

/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const profile2 = getConfiguration('profile2');

/** @type {import("@cucumber/cucumber/lib/configuration").IConfiguration} */
const defaultCucumberConfig = getConfiguration();

const globalConfig = {
  default: defaultCucumberConfig,
  profile1,
  profile2,
}

export default globalConfig;

``` 

I'm aware this is a breaking change, `cucumber-js` could temporaly (until next major) support both syntaxes for a non breaking approach.

### ⛏ Have you considered any alternatives or workarounds?

Yeah, I could use a common js module but if I'm not mistaken I think esm should not be a second class citizien.

### 📚 Any additional context?

Consider [ECMAScript spec](https://tc39.es/ecma262/#sec-exports) as reference

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

davidjgoss:
[...]

PR #2384, opened 2024-03-20
update config to support default definition function

notaphplover:
### 🤔 What's changed?

This PR aims to allow `default` profile to be a function that dynamically generates cucumber profiles

Consider #2009 as reference.

### 📋 Checklist:

  - [ ] I have added an entry to the "Unreleased" section of the [**CHANGELOG**](../blob/main/CHANGELOG.md), linking to this pull request.

----

### DIFF
--- a/src/configuration/from_file.ts
+++ b/src/configuration/from_file.ts
@@ -15,8 +15,21 @@ export async function fromFile(
   file: string,
   profiles: string[] = []
 ): Promise<Partial<IConfiguration>> {
-  const definitions = await loadFile(logger, cwd, file)
-  if (!definitions.default) {
+  let definitions = await loadFile(logger, cwd, file)
+
+  const defaultDefinition: unknown = definitions.default
+
+  if (defaultDefinition) {
+    if (typeof defaultDefinition === 'function') {
+      const definitionsFromDefault = await defaultDefinition()
+
+      definitions = {
+        ...definitions,
+        default: {},
+        ...definitionsFromDefault,
+      }
+    }
+  } else {
     logger.debug('No default profile defined in configuration file')
     definitions.default = {}
   }
--- a/src/configuration/from_file_spec.ts
+++ b/src/configuration/from_file_spec.ts
@@ -98,6 +98,29 @@ describe('fromFile', () => {
       expect(result).to.deep.eq({ paths: ['other/path/*.feature'] })
     })
 
+    it('should work with .mjs with default function', async () => {
+      const { logger, cwd } = await setup(
+        'cucumber.mjs',
+        `export default async function() { 
+          return {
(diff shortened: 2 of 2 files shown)


davidjgoss (commented):
Thanks for raising @notaphplover, this is looking good.

A few initial thoughts from me, if that's okay:

- It's neat that we allow static profile definitions to be done on other exports, but that feels like a bit of a confusing API to me - I think it might be better to keep it simple, so you either a) provide static profiles with exports, or a single default export function to do it dynamically, and not a mixture of the two. What do you think?
- I also think we should avoid adding the `default: {}` in the new code - if the dynamically resolved config has no default key, I still think we want to have that logging happen below

Other than that, I think the next steps here would be:
- Add a `logger.debug` statement when we detect and run a function for config
- Add a new e2e scenario at the end of https://github.com/cucumber/cucumber-js/blob/main/features/profiles.feature
- Add documentation in https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md and/or https://github.com/cucumber/cucumber-js/blob/main/docs/profiles.md

Let me know if you need a steer on anything!

notaphplover:
> Thanks for raising @notaphplover, this is looking good.
> 
> A few initial thoughts from me, if that's okay:
> 
> * It's neat that we allow static profile definitions to be done on other exports when doing the function, but that feels like a bit of a confusing API to me - I think it might be better to keep it simple, so you either a) provide static profiles with exports, or b) a single default export function to do it dynamically, and not a mixture of the two. What do you think?

I totally agree with you.

> * I also think we should avoid adding the `default: {}` in the new code - if the dynamically resolved config has no default key, I still think we want to have that logging happen below

I see, but then in some cases there will be no default profile. You know way better than me the code flows so I assume that's ok :). Sure, I'll add the logging sentence.

> 
> Other than that, I think the next steps here would be:
> 
> * Add a `logger.debug` statement when we detect and run a function for config
> * Add a new e2e scenario at the end of https://github.com/cucumber/cucumber-js/blob/main/features/profiles.feature
> * Add documentation in https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md and/or https://github.com/cucumber/cucumber-js/blob/main/docs/profiles.md

Sure, I'll go for it

> 
> Let me know if you need a steer on anything!
Thank you so much! Sure, I'll do it. I'll try to have some time to continue on this on the weekend.

davidjgoss:
> I see, but then in some cases there will be no default profile. You know way better than me the code flows so I assume that's ok. Sure, I'll add the logging sentence.

Ah you're right, looking at the code again I see that the `else` block with the logging would not be reached in the case of a config-providing function. It would be good to make that happen either way by shifting the code around a bit.

Issue #2392, opened 2024-04-09
Step with result status = pending does not contain an error_message


markstein:
### 🤔 What's the problem you're trying to solve?

By evaluating the runs, it is sometimes helpful to see the detailed error message even for steps that are pending.

This is also how it is implemented in cucumber-jvm.

### ✨ What's your proposed solution?

The error message is also output for steps that have the status pending.

PR #2393, opened 2024-04-09
add error_message for steps that have the status pending #2392

markstein:
### 🤔 What's changed?

By evaluating the runs, it is sometimes helpful to see the detailed error message even for steps that are pending.

The error message is also output for steps that have the status pending #2392 

### 📋 Checklist:

----


davidjgoss (commented):
Thanks for taking the time to do this @markstein.

Can you give an example of user code that you would use and expect this new behaviour to kick in? Generally pending steps are signified with `return 'pending'` (or other equivalent) so there is not an error present, although I suppose with the callback flavour it could work - is that what you had in mind?

See also https://github.com/cucumber/cucumber-js/blob/main/features/pending_steps.feature

davidjgoss:
I'm also curious as to what a "detailed error message" would be in case of pending, given it is supposed to just mean the step isn't implemented yet.

markstein:
@davidjgoss thanks for taking the time for the review.

yes, we use it as a callback variant in the afterStep hook. We use the pending status partly for integration tests that are currently "unstable" due to instability in external systems. However, other sources of error could possibly arise, which would then be recognized based on the error message.

Furthermore, if you look at the Java implementation (which is not necessarily relevant), but the behavior reflects:
https://github.com/cucumber/cucumber-jvm/blob/6549826da802dbf031b89590d2cf4bbdeb36fc3c/cucumber-core/src/main/java/io/cucumber/core/plugin/JsonFormatter.java#L317
image

Based on json result, the error message is then displayed in the html report.

PR #2402, opened 2024-05-06
Add world and context proxies for use in arrow functions

### LINKED ISSUE #2004: Arrow function support with world option

It is such a common problem that users use arrow functions in step definitions, that this is a FAQ item https://github.com/cucumber/cucumber-js/blob/main/docs/faq.md#the-world-instance-isnt-available-in-my-hooks-or-step-definitions

This is likely due to the fact that popular testing libraries such as Jest use of arrow functions https://jestjs.io/docs/api#testname-fn-timeout

The problem is cucumber-js is violating [Principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

I propose an API breaking changes where `world` is passed in as the first parameter of arrow function.

BEFORE
```js
Then(/^the response status is (.*)$/, function (status) {
  assert.equal(this.responseStatus, status)
});
```

BREAKING AFTER
```js
Then(/^the response status is (.*)$/,  (world, status) => {
  assert.equal(world.responseStatus, status)
});
```

This could be introduced as an option with the default as the existing behaviour to allow users try the new behaviour without a breaking semver. Then once stable, release breaking semver and change the default behaviour. This also allows users who wish to upgrade to use option to restore previous default.

No other obvious way to achieve this. The `world` could be added as the last parameter of the arrow function, but I think in the long run that is more confusing.

### 📚 Any additional context?

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/feature_request.md)*

davidjgoss:
### 🤔 What's changed?

This PR adds two new public exports:

- `world` which you can reference from a step definition or hook to get the World
- `context` which you can reference from a `BeforeAll` or `AfterAll` hook to get the context containing the world parameters

I've added this to the docs, but assuming this is successful with early adopters I will raise a follow up PR to switch all our code examples over to use it too, and mark it stable (currently it's beta).

This provides a solution for the long-standing conflict between wanting to access the "World" instance which has always been bound to `this`, and wanting to use modern JavaScript syntax with arrow functions that don't honour such a binding.

You can now write step definitions with arrow functions and use the World like this:

```javascript
const { Given, world } = require('@cucumber/cucumber')

Given('a step', () => {
  world.log(world.parameters.foo)
})
```

Fixes #2004.

The exported `world` object is a `Proxy` which dynamically forwards to the World for the currently-executing test case, which in turn is provided by a Node.js `AsyncLocalStorage` which works very similarly to React context in that we can have a piece of state scoped to the running of some code, and reachable from within that code without passing directly. 

There are a couple of things to bear in mind with this approach:

- The proxy is very comprehensive, but it's possible if someone is doing something _really_ weird with their World they could have issues. But then, they always have the old pattern to fall back on.
- If at a future point we target runtimes besides Node.js, we may have to provide alternate implementations - as such, the use of `AsyncLocalStorage` here is behind a generic facade and does not leak externally.

### 📋 Checklist:

----

### DIFF
--- a/docs/transpiling.md
+++ b/docs/transpiling.md
@@ -56,6 +56,10 @@ You then just need to specify the `import` option as above for your support code
 
 (This approach is no longer recommended, and you might see a warning from Node.js telling you so.)
 
+### tsconfig-paths
+
+It's not unusual for people to use some path remapping and [`tsconfig-paths`](https://www.npmjs.com/package/tsconfig-paths) as part of their TypeScript setup. See [this open issue](https://github.com/dividab/tsconfig-paths/issues/243) regarding ESM support in that library.
+
 ## Source maps
 
 Source maps are used to ensure accurate source references and stack traces in Cucumber's reporting, by giving traceability from a transpiled piece of code back to the original source code.


mpkorstanje:
One last thought. To make API evolution easier, maybe don't put methods like `log` and `attach` on world. The world belongs to the user so to say.

And if world is backed by a thread local and passed to another thread (some async call back) what happens?

PR #2547, opened 2025-05-04
AIC-3548-dev

Aryan2180:
MAke the report link global and it will console at the end of the execution after the RootCause and Failures


Aryan2180:
wrong pr

PR #2596, opened 2025-07-22
fix: properly handle ES module default exports in configuration files

lackita:
When importing ES module configuration files, the import() function returns an object with a 'default' property containing the exported value. This fix extracts the default export correctly to prevent double-nesting of configuration objects that would make them inaccessible during parsing.

Fixes issue where ES module configuration files with export default syntax would not be properly parsed, causing step definitions and other configuration to be ignored.

🤖 Generated with [Claude Code](https://claude.ai/code)

### 📋 Checklist:

- [X] I've changed the behaviour of the code
  - [X] I have added/updated tests to cover my changes.

### DIFF
--- a/features/esm.feature
+++ b/features/esm.feature
@@ -125,3 +125,57 @@ Feature: ES modules support
       """
         Error: Cucumber expected a CommonJS module
       """
+
+  @esm
+  Scenario: ES module configuration file with export default should load step definitions
+    Given a file named "features/a.feature" with:
+      """
+      Feature:
+        Scenario: one
+          Given a step that should be found
+      """
+    And a file named "features/step_definitions/cucumber_steps.js" with:
+      """
+      import {Given} from '@cucumber/cucumber'
+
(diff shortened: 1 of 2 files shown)


davidjgoss:
Thanks for taking the time to contribute @lackita.

However, the changes here have no real effect, they just introduce another variable in a couple of places.

The way that configuration files work with ESM is not a bug - your default profile should be declared with a default export (e.g. `export default`) and any other profiles should be declared with named exports (e.g. `export const otherProfile`). You can see this in the difference between the ESM and CommonJS examples in the [documentation for configuration](https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md), but to be fair this could be clearer in the [documentation for profiles](https://github.com/cucumber/cucumber-js/blob/main/docs/profiles.md) - I'd be happy to accept a pull request for adding a decent ESM example there if you can.

If you have a use case that you can't get to work, please do raise an issue with a detailed example. For now I'm going to close this PR.

Issue #2303, opened 2023-07-25
Improve execution time by executing on multiple machines


MiladSadinam:
### 🤔 What's the problem you're trying to solve?

We love cucumber and use it for end-to-end testing our UI. As the tests include the UI, backend and database, they are quite "heavy" and are taking more and more time. We would like to decrease the execution time, especially as they are used in our pull-request CI/CD. 

### ✨ What's your proposed solution?

Another quick way to speed up tests is to split them across multiple machines. I believe Cucumber-Js currently does not support this, correct? Or is there any third-party package I did not find? 

We would love that feature. Nowadays, hardware is getting more cheap, and you can so easily speed up the workflow by paying little more money.

### ⛏ Have you considered any alternatives or workarounds?

Cucumber-js supports the `parallel` parameter, but our pull-request provider (Atlassian Bitbucket) only supports a limited number of (not very strong) machines, so we currently cannot go beyond `2` for the `parallel` parameter.

davidjgoss:
@MiladSadinam thanks for raising, you'r right that it's currently not supported. `parallel` just uses multiple worker threads but on the same machine which for reasons you've noted has limitations.

From what you've written, I think the kind of thing you're looking for is more like [Playwright's sharding feature](https://playwright.dev/docs/test-parallel#shard-tests-between-multiple-machines), where an independent run would execute a subset of test cases based on an index and count, e.g.:

```shell
# machine 1
npx cucumber-js --shard 1/3

# machine 2
npx cucumber-js --shard 2/3

# machine 3
npx cucumber-js --shard 3/3
```

I'd definitely be supportive of doing this as a relatively low-effort way to give people another lever for performance.

- - -

Some thoughts about how this should work (more for internal audience):

- When do we run the logic to get the subset of test cases for this shard? We should be able to do it once we have derived the pickles. Presumably then we wouldn't emit `testCase` messages for the ones we aren't going to run. I think we probably couldn't avoid emitting the `pickle` message if we wanted to (without a significant rework), since that is done by Gherkin before any of the Cucumber processing kicks in.
- Once we release this, people are (reasonably) going to want to be able to e.g. concatenate the messages from the multiple sharded runs and produce a single HTML report. I doubt we would support this right away, but we should avoid doing anything that would block it later. We've already taken some small steps towards this with e.g. making `testRunStarted` messages have a unique id, so it's not totally unrealistic.
- The `parallel` option should still work in the context of one shard.
[...]

PR #2600, opened 2025-07-27
Add support for execution sharding

agologan:
### 🤔 What's changed?

- Adds new flag `--shard 1/3` to allow for execution sharding similar to https://playwright.dev/docs/test-sharding

Tackles #2303

- :book: Documentation (improvements without changing code)

Feature is implemented as plugin on event `pickles:filter` and will run between filter and order plugins.
As such `defined` order is not affected by sharding and `random` order will only scramble tests in shard.

Alternatively, a new event `pickles:shard` could be introduced to be executed after `pickles:order` which would allow a global random seed. This will require a documentation change to warn users they should the same seed across shards.

### 📋 Checklist:

----


Tallyb:
From looking at the code, I coudl not tell how the pickles are sorted. If this is a random sort, based on how they are read from the file system, this will lead to errors. You must sort all pickles in a unique way. As a side note, since you use modulo, sorting by file size and then alphabetically will give better performance compared to alphabetically only.

davidjgoss:
@Tallyb can you elaborate on what errors you are foreseeing? Are you able to reproduce them in a sample project?

This PR doesn't deal with any sorting - that is handled in a later phase, after filtering.

Tallyb:
let's take a simple example: 
Assume we have tests A,B,C,D, E, F, G and we run them on shard = 3. 
Each shard is running in a different machine, so it is completely unaware of what is happening on the other machines. 

Machine 1: tests are read in the following order - A, B, C, D, E, F, G. tests taken - A, D, G. (positions: 1, 4, 7)
Machine 2: tests are read in the following order - B, A, C, D, E, F, G. tests taken: A, E, (2, 5)
Machine 2: tests are read in the following order - B, C, A, D, E, F, G. tests taken: A, F, (3, 6)
As you can see, A is tested 3 times, while B and C are not tested at all. to make sure everything is tested, you must guarantee the same order in which tests are being read.

agologan:
The code in all instances does 3 things in this order:
- discovers paths
- filters the paths
- applies ordering (if not default: `defined`)

Can see this is [load_sources.ts](https://github.com/cucumber/cucumber-js/blob/main/src/api/load_sources.ts) or [run_cucumber.ts](https://github.com/cucumber/cucumber-js/blob/main/src/api/run_cucumber.ts)

Sharding is applied after discovery but before further ordering.

So discovery is responsible for the `defined` order and that can be seen in [paths.ts](https://github.com/cucumber/cucumber-js/blob/main/src/paths/paths.ts)

Getting back to your example, all 3 machines run the exact same code to discover 
the paths which as the [docs](https://github.com/cucumber/cucumber-js/blob/main/docs/filtering.md#order) put it "roughly means alphabetical order of file path 
followed by sequential order within each file".

So all 3 machines will discover the tests in an order which roughly resembles 
A, B, C, D, E, F, G order, and is the same across them.

Then all 3 machines will apply the same filtering logic, which would leave them 
with the same sequence, to which sharding is applied.

Picking nth test on each shard will provide uniform distribution of all tests afterwards.

As a last step if a random order is used, those left on each shard are reshuffled.

davidjgoss:
Thanks for laying that out @agologan.

As you say, the initial order of sources and pickles is deterministic given the same configuration, so I don't believe we have a problem here.

Tallyb:
Ok, as long as it is deterministic, there is no issue. 
You can use my tip for improving performance by sorting by size (assuming size ~= execution time).

PR #2642, opened 2025-09-06
Enhance transpiling.md with tsx setup details

Izhaki:
### 🤔 What's changed?

Added a small callout about using `tsx` instead of `ts-node`.

`tsx` is faster and its setup is less complex as it need nothing typescript.

- :book: Documentation (improvements without changing code)


davidjgoss (approved):
Thanks for doing this @Izhaki. I've taken it a step further and emphasised tsx as the primary suggestion, for all the reasons you say.

PR #2763, opened 2026-02-28
chore(deps): replace chalk with picocolors

roli-lpci:
## Summary

- Replace `chalk` (4 transitive deps: chalk, ansi-styles, supports-color's dep has-flag) with `picocolors` (zero deps) for terminal color output
- Map `new chalk.Instance(support)` to `picocolors.createColors(!!support)` — supports-color returns `false` for no-support and `{level: 1|2|3}` for support, so the boolean coercion preserves the same enable/disable semantics
- All colors used (red, green, yellow, cyan, gray) are basic ANSI codes, so the level-to-boolean mapping has no visual impact
- Normalize `chalk.grey` to `picocolors.gray` (both emit ANSI code 90)
- Eliminate the manual identity-function fallback branch — `createColors(false)` handles this automatically

## Changes

- `src/formatter/get_color_fns.ts` — Replace chalk.Instance + bind pattern with picocolors.createColors, simplify from 46 lines to 30
- `features/support/warn_user_about_enabling_developer_mode.ts` — Replace `chalk.red()` with `pc.red()`
- `package.json` — Swap `chalk` for `picocolors` in dependencies
- `package-lock.json` — Regenerated

## Notes

- `supports-color` is intentionally kept — it is used by `src/publish/publish_plugin.ts` outside of the chalk context
- Prior chalk v5 upgrade PR (#1914) was closed — chalk v5 is ESM-only while cucumber-js has CJS exports. Picocolors supports both CJS and ESM.

## Testing

- 408 unit tests passing
- TypeScript build clean
- Prettier clean
- ESLint clean
- dependency-lint clean


davidjgoss (commented):
Thanks for doing this @roli-lpci.

Since then though, we have landed https://github.com/cucumber/cucumber-js/pull/2769 which means we can take this a step further and remove the libraries altogether in favour of [`styleText` from `node:util`](https://nodejs.org/api/util.html#utilstyletextformat-text-options). If passing in the stream as an option, we wouldn't even need the `supports-color` check any more either.

Would you be up for pivoting the PR to do this change instead?

davidjgoss:
Taking https://github.com/cucumber/cucumber-js/pull/2828 forward to lose the dependency entirely - thanks again for the effort though.

Issue #2800, opened 2026-05-04
Dependency tree size


m-basov:
### 🤔 What's the problem you've observed?

Hi! Thanks for creating and maintaining such helpful library! 

I was recently inspecting my node_modules tree and noticed that cucumber adds a significant amount of transitive dependencies. While npm has a rich ecosystem of packages that helps building projects faster, I think such fundamental package as cucumber should be on a more conservative side and use only packages that provide non-trivial/hard to maintain functionality. It is installed millions of times each week in both CI and developer environments so small savings quickly add up! But before I spend more time on this I wanted to ask if there is any interest in moving in that direction? 
I understand that some people prefer building on top of small well-maintained packages instead of rewriting things from scratch so I am not going to push in the other direction if that's the preference. Also, the suggestions I am making is increasing reliance on built-in Node.js libraries so if there are plans to be platform independent I think it makes sense to stay as is.

Here is where cucumber stands now:

<img width="402" height="512" alt="Current dependency tree" src="https://github.com/user-attachments/assets/fa026a0c-19e8-4576-baee-21c7a5b91680" />

### ✨ Do you have a proposal for making it better?

I checked if there are any low hanging fruits and turns out there are quite many! Here is [my branch](https://github.com/cucumber/cucumber-js/compare/main...m-basov:cucumber-js:refactor/trim-down-dependency-tree) where I replaced:
- [capital-case](https://github.com/cucumber/cucumber-js/commit/1f7985c582526b08210bfb5697b279b0bf4d603b)
- [read-package-up](https://github.com/cucumber/cucumber-js/commit/2e36f403293b98cf99816e24414deb42f0f9c2da)
- [mz](https://github.com/cucumber/cucumber-js/commit/014c1331309189c3982074e296ba3c4e6cf416aa)
- [glob](https://github.com/cucumber/cucumber-js/commit/7569d190f444c92c6ca25d77a40bb8374e5d8e75)
- [chalk](https://github.com/cucumber/cucumber-js/commit/028a2c0602b40ca98ab24856a33953bc65f0b63e)

with either built-in Node.js libraries or small functions that do exactly the thing that previously was done by one of the deleted packages and the results are quite exciting -41 transitive dependency!

<img width="402" height="481" alt="Trimmed down dependency tree" src="https://github.com/user-attachments/assets/529a910e-758b-4e8a-aadf-a0af59bb0062" />

[...]

PR #2801, opened 2026-05-06
Replace `capital-case` with custom implementation

m-basov:
### 🤔 What's changed?

Replace [capital-case](https://npmx.dev/package/capital-case) dependency (deprecated) with a simpler custom implementation which decreases dependency tree size by 5 packages total.

I've described my motivation in #2800. For `capital-case` specifically we need to transform a predefined list of properties which allows us to ignore complex edge cases and just use a naive implementation.

I am specifically curious about testing approach to this kind of changes. The change I made already has tests so I haven't added any but I am happy to add more if you think it would be valuable.

### 📋 Checklist:

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,6 @@
         "@cucumber/pretty-formatter": "3.3.0",
         "@cucumber/tag-expressions": "9.1.0",
         "assertion-error-formatter": "^3.0.0",
-        "capital-case": "^1.0.4",
         "chalk": "^4.1.2",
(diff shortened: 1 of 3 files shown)


davidjgoss (approved):
Great, thanks for doing this.

> I am specifically curious about testing approach to this kind of changes. The change I made already has tests so I haven't added any but I am happy to add more if you think it would be valuable.

In this case the existing coverage is good, but we should look at it case by case for future ones. Acceptance-level coverage is ideal.

PR #2809, opened 2026-05-14
Use built-in fs.glob

m-basov:
### 🤔 What's changed?

Use [glob](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisesglobpattern-options) from `node:fs` instead of [glob package](https://npmx.dev/package/glob).

My motivation is described in #2800. Starting from Node 22 there is a built-in glob which allows us to drop 6 packages from prod dependencies (glob is still installed in dev transitively) and save ~1.6MB. Built-in glob returns an async iterator with relative paths (unless `withFileTypes` is provided) so I've updated paths.ts to follow the previous behaviour.  

I used [Array.fromAsync](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync#browser_compatibility) which is available in Node 22 but it surfaced two issues:
- Current tsconfig.json is slightly outdated so I borrowed the minimum amount of changes needed from https://github.com/tsconfig/bases/blob/main/bases/node22.json but we may consider using updating the full config eventually?
- `@types/lodash` is installed transitively and is very old, using it with `esnext.Array` lib causes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/15331 so I had to run `npm upgrade @types/lodash`.

### 📋 Checklist:

### DIFF
--- a/src/paths/paths.ts
+++ b/src/paths/paths.ts
@@ -1,5 +1,6 @@
 import path from 'node:path'
-import { glob } from 'glob'
+// @ts-expect-error -- Requires Node 22
+import { glob } from 'node:fs/promises'
 import fs from 'mz/fs'
 import { ILogger } from '../environment'
 import { ISourcesCoordinates, ISupportCodeCoordinates } from '../api'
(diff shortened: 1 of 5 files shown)


davidjgoss:
@m-basov with https://github.com/cucumber/cucumber-js/pull/2818 now merged you should be able to rebase and get this green.

src/paths/paths.ts
```diff
@@ -54,17 +53,14 @@ async function expandPaths(
...
-        matches.map(async (match) => {
+      const matches: string[] = await Array.fromAsync(
+        glob(unexpandedPath, { cwd })
+      )
+      const expanded: string[][] = await Promise.all(
+        matches.map(async (matchRelative) => {
+          const match = path.resolve(cwd, matchRelative)
           if (path.extname(match) === '') {
-            return glob(`${match}/**/*${defaultExtension}`, {
-              windowsPathsNoEscape: true,
```
m-basov, line 66: I was considering writing a windows path test to ensure that this works as expected in the built-in glob but I think it should be covered by path.join (because it's platform specific) https://github.com/cucumber/cucumber-js/blob/bb2b30304b4beb691ac1f3bccf435104973f2500/src/paths/paths_spec.ts#L24 when we run tests on windows. Having said that windows-like globs seem to work just fine on linux so I can add an explicit test case if needed.

davidjgoss: Yeah we would see a failing test in CI where you have pointed to if we had a Windows problem, so all good on that front.

davidjgoss (approved):

PR #2813, opened 2026-05-18
Surface Error.cause in formatted error output

### LINKED ISSUE #2515: Cucumber does not support Error cause.
### 👓 What did you see?

Error cause is lost in format_error and cucumber/pretty-formatter does not print causes.

At our project we log errors with causes, but the cause is filtered out by cucumber.
I noticed this happens in format_error where a new error is built in the exception property. 

I manually added in my node_modules for cucumber in format_error.js (just to test) the cause. Next, I still saw that the cause was ignored, hence it being emitted on the envelope.

The reason we log causes is because we wrapped cucumber to work with our own testing framework. We work with the cucumber/api 

We expect the cause to be emitted by cucumber, and if possible, also be formatted by the pretty-formatter.

node v18.20.2
cucumber ^v11.1

try {
  throw new Error('Original cause error')
}
catch(e) {
   throw new Error('message', { cause: e })
}

### 📚 Any additional context?

_No response_

rkdfx:
### 🤔 What's changed?

If a thrown error has a `cause`, we now keep it and add it to the formatted output as `Caused by: <type>:<message>`.

Closes #2515.

### 📋 Checklist:

----


davidjgoss (changes requested):
Thanks for doing this @rkdfx.

What I think we're missing here is:

- The ability to handle multiple nesting levels, not just one `cause` - we'd have to have a brake at say 10 levels, and for circular references
- The stack frames of the nested error(s), not just the top-level one

So with your example from the feature file, I'd be looking for output like:

```
Error: wrapped failure
    at World.<anonymous> (/Users/davidgoss/Projects/cucumber-js-harness/features/steps.ts:41:15)
Caused by: Error: original cause
    at World.<anonymous> (/Users/davidgoss/Projects/cucumber-js-harness/features/steps.ts:39:15)
```

If you'd be able to make those changes that would be great, but if not I'm happy to merge this as-is and iterate - let me know.

davidjgoss (approved):

PR #2845, opened 2026-06-06
Replace is-stream dependency

m-basov:
### 🤔 What's changed?

Replace [is-stream](https://github.com/sindresorhus/is-stream) package with built-in [isReadable](https://nodejs.org/docs/latest-v22.x/api/stream.html#streamisreadablestream).

Continuation of #2800. These implementations have slight implementation differences but unless someone was trying to spoof a readable object instead of passing the actual readable they should work as expected. Both check that the passed object is a stream, it's open, and readable.
- [node.js](https://github.com/nodejs/node/blob/7b1b943359517852235ebda83410ca003ffa3df9/lib/internal/streams/utils.js#L173)
- [is-stream](https://github.com/sindresorhus/is-stream/blob/ab06c4acc9dce4dcadc9dfc6416e1be2c836862d/index.js#L19)

I've had to swap the order of string and readable checks because the built-in utility only accepts `Readable` and `ReadableStream` objects.

### 📋 Checklist:


exports/root/report.api.md
```diff
@@ -13,7 +13,7 @@ import type { JsonObject } from 'type-fest';
 import * as messages from '@cucumber/messages';
 import { ParameterType } from '@cucumber/cucumber-expressions';
 import { ParameterTypeRegistry } from '@cucumber/cucumber-expressions';
-import type { Readable } from 'node:stream';
+import { Readable } from 'node:stream';
```
m-basov, line 16: I am not sure how the api-exports work. This happens because I import `isReadable` from `node:stream` now so there is no more import ellision.

davidjgoss (approved):

PR #2846, opened 2026-06-06
Replace mkdirp dependency

m-basov:
### 🤔 What's changed?

Use built-in [mkdir](https://nodejs.org/docs/latest-v22.x/api/fs.html#fspromisesmkdirpath-options) instead of [mkdirp](https://github.com/isaacs/node-mkdirp) package.

Continuation of #2800. According to mkdirp's own [comparison](https://github.com/isaacs/node-mkdirp#use-fsmkdirpath-recursive-true-cb-if) should be used only if a handful of specific scenarios and from what I can tell cucumber doesn't care about any of those.

This is indirectly tested by using cucumber as a test runner for its own tests. But might be worth adding a unit test?

### 📋 Checklist:

### DIFF
--- a/exports/root/report.api.md
+++ b/exports/root/report.api.md
@@ -41,8 +41,7 @@ export const BeforeAll: ((code: TestRunHookFunction) => void) & ((options: IDefi
 export const BeforeStep: (<WorldType = IWorld<any>>(code: TestStepHookFunction<WorldType>) => void) & (<WorldType = IWorld<any>>(tags: string, code: TestStepHookFunction<WorldType>) => void) & (<WorldType = IWorld<any>>(options: IDefineTestStepHookOptions, code: TestStepHookFunction<WorldType>) => void);
 
 // @beta
-const context_2: IContext<any>;
-export { context_2 as context }
+export const context: IContext<any>;
 
 // @public (undocumented)
 export class DataTable {
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
     "moduleResolution": "NodeNext",
     "resolveJsonModule": true,
     "typeRoots": ["./node_modules/@types", "./src/types"],
+    "types": ["node"],
     "declaration": true,
     "inlineSourceMap": true,
     "inlineSources": true,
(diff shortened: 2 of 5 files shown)


tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "moduleResolution": "NodeNext",
     "resolveJsonModule": true,
     "typeRoots": ["./node_modules/@types", "./src/types"],
+    "types": ["node"],
```
m-basov, line 10: This is one of the breaking changes of TS 6.0 https://devblogs.microsoft.com/typescript/announcing-typescript-6-0/#up-front-adjustments but somehow cucumber was able to work without specifying types. However, removing mkdirp breaks this in a catastrophic way for an unknown to me reason.

exports/root/report.api.md
```diff
@@ -41,8 +41,7 @@ export const BeforeAll: ((code: TestRunHookFunction) => void) & ((options: IDefi
 export const BeforeStep: (<WorldType = IWorld<any>>(code: TestStepHookFunction<WorldType>) => void) & (<WorldType = IWorld<any>>(tags: string, code: TestStepHookFunction<WorldType>) => void) & (<WorldType = IWorld<any>>(options: IDefineTestStepHookOptions, code: TestStepHookFunction<WorldType>) => void);
 
 // @beta
-const context_2: IContext<any>;
-export { context_2 as context }
+export const context: IContext<any>;
```
m-basov, line 44: I have no idea why this happens. Does removing/adding imports to internal modules counts as an API change?

davidjgoss: I've spent some time looking at this and can only conclude that _something_ about `mkdirp`, in the context of the wider dependency tree, was making it think there was another symbol called `context` floating around somewhere that it had to avoid a collision with, and now with `mkdirp` gone it doesn't any more. Very weird.

davidjgoss: It's looking like `mocha` which adds `context` as a global. Adding the explicit `types` array to the tsconfig may have bounced this out.

tsconfig.json
davidjgoss: I think the node types were getting silently brought into scope by `mkdirp` which reuses and extends some node types, and so we got away with it before.

davidjgoss (approved):

PR #2860, opened 2026-06-22
Allow configuration via environment variables

### LINKED ISSUE #2347: allow configuration via environment variables
### 🤔 What's the problem you've observed?

Cucumber can be configured via the CLI and files, but not via environment variables (except the `publish` functionality which does support some).

This means users that configure most of their other stuff with environment variables often have a sort of frankenstein setup, with some stuff living in the Cucumber config (maybe in profiles) and other somewhat related stuff sitting elsewhere (e.g. in `.env` files).

### ✨ Do you have a proposal for making it better?

Support configuring Cucumber via environment variables.

When resolving configuration, each option should be sought from environment variables by converting its key to uppercase-with-underscores format, prefixing with `CUCUMBER_`, and then checking the environment for that key. So for example, `retryTagFilter` would be expressed as `CUCUMBER_RETRY_TAG_FILTER` and so on.

TBA on how things like booleans and arrays should be expressed and parsed. We should follow conventions set by other tools if they exist.

Environment variables should take precedence over the configuration file and CLI. All options should maintain their current merge vs overwrite behaviour.

Profile-level options are out of scope.

### 📚 Any additional context?

See conversation on https://github.com/cucumber/cucumber-js/issues/2346.

----

*This text was originally generated from a [template](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates), then edited by hand. [You can modify the template here.](https://github.com/cucumber/.github/edit/main/.github/ISSUE_TEMPLATE/developer_experience.md)*

youdie006:
### 🤔 What's changed?

Each configuration option can now be set via an environment variable (#2347). An option's key is converted to `CUCUMBER_OPTION_<SCREAMING_SNAKE>` — e.g. `retryTagFilter` → `CUCUMBER_OPTION_RETRY_TAG_FILTER` — following @davidjgoss's design in the issue.

- **Precedence**: CLI > environment > config file > default (env slots into `mergeConfigurations` between the config file and CLI), preserving the existing merge-vs-overwrite semantics (e.g. `CUCUMBER_OPTION_TAGS` and `--tags` combine into `(@a) and (@b)`).
- Values are parsed as JSON where possible (booleans / numbers / arrays) and validated against the config schema; unknown `CUCUMBER_OPTION_*` vars are ignored. Profile-level options are out of scope, per the issue.
- New internal `fromEnv` module; the public API surface is unchanged (verified by api-extractor).

Fixes #2347 — lets every configuration option be provided via the environment (useful in CI), which previously wasn't possible for most options.

The precedence placement (env between config file and CLI) and the `CUCUMBER_OPTION_TAGS` merge semantics — happy to adjust if you'd prefer different behaviour. Quick manual check: `CUCUMBER_OPTION_DRY_RUN=true` skips scenarios; `CUCUMBER_OPTION_TAGS='@wip'` with `--tags 'not @wip'` resolves to `(@wip) and (not @wip)`.

### 📋 Checklist:

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -84,12 +84,37 @@ export default {
 } satisfies Partial<IConfiguration>
 ```
 
+## Environment variables
+
+You can also set configuration options via environment variables. For each option, take its key, convert it to uppercase-with-underscores, and prefix it with `CUCUMBER_OPTION_`. For example, `retryTagFilter` becomes `CUCUMBER_OPTION_RETRY_TAG_FILTER`:
+
+```shell
+CUCUMBER_OPTION_PARALLEL=2 CUCUMBER_OPTION_RETRY_TAG_FILTER='@flaky' cucumber-js
+```
+
+Values are parsed as JSON where possible, so booleans, numbers, arrays and objects can be expressed, while anything that isn't valid JSON (like a tag expression) is kept as a plain string:
+
+```shell
+# boolean
+CUCUMBER_OPTION_DRY_RUN=true
+# number
+CUCUMBER_OPTION_PARALLEL=2
+# string
+CUCUMBER_OPTION_TAGS='@foo and @bar'
+# array
+CUCUMBER_OPTION_FORMAT='["html:cucumber-report.html"]'
+# object
+CUCUMBER_OPTION_WORLD_PARAMETERS='{"baseUrl":"https://example.com"}'
+```
+
+Environment variables take precedence over the configuration file and defaults, but not over the CLI, which always wins because it's passed directly in the command. Each option keeps its usual merge-vs-overwrite behaviour, so for example tag expressions from the environment and the CLI are combined rather than replaced. Profile-level options aren't supported via environment variables.
+
 ## Options
 
-These options can be used in a configuration file (see [above](#files)) or on the [CLI](./cli.md), or both.
+These options can be used in a configuration file (see [above](#files)), via [environment variables](#environment-variables) or on the [CLI](./cli.md), or any combination.
 
 - Where options are repeatable, they are appended/merged if provided more than once.
-- Where options aren't repeatable, the CLI takes precedence over a configuration file.
+- Where options aren't repeatable, the order of precedence (highest first) is CLI, environment variable, configuration file.
 
 | Name              | Type       | Repeatable | CLI Option                | Description                                                                                                        | Default |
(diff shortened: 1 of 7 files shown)


davidjgoss (changes requested):
Thanks for raising this @youdie006. It looks good at a glance, but please can you follow the standard pull request template including the CoC acknowledgement:
https://raw.githubusercontent.com/cucumber/.github/refs/heads/main/.github/PULL_REQUEST_TEMPLATE.md

youdie006:
Thanks @davidjgoss! I've updated the description to the standard template with the Code of Conduct acknowledgement, and pointed the CHANGELOG entry at this PR. Ready for another look when you have a moment.

docs/configuration.md
```diff
@@ -84,12 +84,37 @@ export default {
...
 
+## Environment variables
+
+You can also set configuration options via environment variables. For each option, take its key, convert it to uppercase-with-underscores, and prefix it with `CUCUMBER_OPTION_`. For example, `retryTagFilter` becomes `CUCUMBER_OPTION_RETRY_TAG_FILTER`:
+
+```shell
+CUCUMBER_OPTION_PARALLEL=2 CUCUMBER_OPTION_RETRY_TAG_FILTER='@flaky' cucumber-js
+```
+
+Values are parsed as JSON where possible, so booleans, numbers, arrays and objects can be expressed, while anything that isn't valid JSON (like a tag expression) is kept as a plain string:
```
davidjgoss, line 95: This is an implementation detail and we could do without having it in the user-facing documentation.

src/configuration/from_env.ts
davidjgoss, line 1: Looking at this now, I think my steer to have `CUCUMBER_OPTION` be the prefix we look for was a mistake and it's led us to a more complex implementation than is needed.

First, there are no collisions with other env vars, and given we also support `CUCUMBER_PUBLISH_URL` and `CUCUMBER_PUBLISH_TOKEN`, having the extra prefix causes fragmentation from a user perspective. So I think just e.g. `CUCUMBER_RETRY_TAG_FILTER` is fine.

Second, we already have the list of option names to check (from the default flag config object), so we can have fewer loops and less code by simply iterating over those keys, and for each one forming the env var key and checking `env` for it - no need to iterate over the whole env object.

davidjgoss (commented):
Thanks again. A couple of suggestions to improve.

PR #2874, opened 2026-07-09
Use fake timers instead of real setTimeout waits

nikolauspschuetz:
### 🤔 What's changed?

The `stopwatch` specs waited on real `setTimeout` calls — 1200 + 200 + 200 + 200 + 200 ms — so running the file spent roughly two seconds just sleeping, then checked the measured duration with a `closeTo(…, 50)` tolerance.

I swapped the real waits for `@sinonjs/fake-timers` and advance the clock with `clock.tick(…)`. I reused the `FakeTimers.withGlobal(timeMethods).install()` setup that already exists in `test_case_runner_spec.ts` and `assemble_test_cases_spec.ts`, because the stopwatch reads the clock through the shared `../time` methods object, so faking that object is enough. The four duration tests are now synchronous and assert exact values (`to.eq(1200)`), and the file runs in ~12ms. I left the `returns a timestamp close to now` test on the real clock since it's already instant.

Two things bugged me about these specs: they're slow for no real reason (sleeping in a unit test), and the `closeTo(…, 50)` tolerance is a latent flake — under CI load the event loop can drift past 50ms and fail an assertion that has nothing to do with an actual regression. Faking the clock removes both problems at once.

I scoped the fake clock to a nested `describe('duration')` so the `timestamp` test keeps the real clock. If you'd prefer I fake the whole file and pin `timestamp` against a fixed `now` instead, I'm happy to change it.

### 📋 Checklist:

- [x] I agree to respect and uphold the Cucumber Community Code of Conduct
- [ ] I've changed the behaviour of the code — no, this is test-only; behaviour is unchanged and the assertions are just tightened
- [ ] My change requires a change to the documentation
- [ ] Users should know about my change — no, it's an internal test change with nothing user-facing, so no CHANGELOG entry

### DIFF
--- a/src/runtime/stopwatch_spec.ts
+++ b/src/runtime/stopwatch_spec.ts
@@ -1,40 +1,54 @@
 import { TimeConversion } from '@cucumber/messages'
+import FakeTimers, { type InstalledClock } from '@sinonjs/fake-timers'
 import { expect } from 'chai'
-import { describe, it } from 'mocha'
+import { afterEach, beforeEach, describe, it } from 'mocha'
+import timeMethods from '../time'
 import { create, timestamp } from './stopwatch'
 
 describe('stopwatch', () => {
-  it('returns a duration between the start and stop', async () => {
-    const stopwatch = create()
-    stopwatch.start()
-    await new Promise((resolve) => setTimeout(resolve, 1200))
-    stopwatch.stop()
-    expect(TimeConversion.durationToMilliseconds(stopwatch.duration())).to.be.closeTo(1200, 50)
-  })
+  describe('duration', () => {
+    let clock: InstalledClock
 
-  it('accounts for an initial duration', async () => {
-    const stopwatch = create(TimeConversion.millisecondsToDuration(300))
-    stopwatch.start()
-    await new Promise((resolve) => setTimeout(resolve, 200))
-    stopwatch.stop()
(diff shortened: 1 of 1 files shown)


src/runtime/stopwatch_spec.ts
```diff
@@ -1,40 +1,59 @@
...
 
-  it('accounts for an initial duration', async () => {
-    const stopwatch = create(TimeConversion.millisecondsToDuration(300))
-    stopwatch.start()
-    await new Promise((resolve) => setTimeout(resolve, 200))
-    stopwatch.stop()
-    expect(TimeConversion.durationToMilliseconds(stopwatch.duration())).to.be.closeTo(500, 50)
-  })
+    beforeEach(() => {
+      // Fake the timing builtins the stopwatch reads (performance/Date on the
```
davidjgoss: We don't need to document what sinon fake timers do here - please can we remove this comment.

davidjgoss (changes requested):
Thanks, this is a good improvement. Just a small request for changes.

src/runtime/stopwatch_spec.ts
nikolauspschuetz: Good call — removed. Thanks for the review!

davidjgoss (approved):

Issue #2907, opened 2026-07-31
Parallel runtime hangs indefinitely when 0 scenarios match (regression since 13.0.0 worker-thread rewrite)


hbl18:
### 👓 What did you see?

When running `cucumber-js` with `parallel` set to a value like `2` and the configured `paths`/`tags` result in **zero matching scenarios**, the process hangs forever. It never prints a summary, never emits `TestRunFinished`, and never exits — it has to be killed manually (e.g. `SIGKILL`/CI job timeout).

This happens on both `13.0.0` and the current latest `13.2.0`.

With `parallel: 0` (serial mode), the same configuration correctly prints `0 scenarios` / `0 steps` and exits immediately with code 0.

### ✅ What did you expect to see?

Same as serial mode: the run should finish immediately, print `0 scenarios`, and exit with code 0 (no hang), since there is no work to distribute to worker threads.

### 📦 Which tool/library version are you using?

- `@cucumber/cucumber`: `13.0.0` and `13.2.0` (reproduced on both)
- Also confirmed **not** present in `12.9.0` (exits cleanly with 0 scenarios in parallel mode) — this is a regression introduced by the parallel runtime rewrite in 13.0.0 (#2710, worker-thread based coordinator).
- Node.js: v24.18.1
- OS: macOS (Darwin)

### 🔬 How could we reproduce it?

1. Create a minimal project:

```
mkdir repro && cd repro
mkdir features
```

2. Add  `features/test.feature`:

```
@sometag
Feature: Sample
  Scenario: A scenario
    Given a step
```

3. Add `cucumber.mjs`:

```js
export default {
  paths: [],
  tags: '@nonexistent', // matches nothing in features/test.feature
  format: ['summary'],
  parallel: 2,
};
```

4. Install and run:

```
npm install @cucumber/cucumber@13.2.0
npx cucumber-js
```

**Result:** process hangs indefinitely (no output, no exit).

5. For comparison, change `parallel: 2` to `parallel: 0` in the same config → process exits immediately with:
```
0 scenarios
0 steps
0m 0.0s (0m 0.0s executing your code)
```

6. For comparison with a pre-13 version:
```
npm install @cucumber/cucumber@12.9.0
npx cucumber-js
```
→ exits immediately with `0 scenarios` even with `parallel: 2`.

### 📚 Any additional context?

[...]

PR #2908, opened 2026-07-31
fixes parallel runtime hanging with 0 scenarios

hbl18:
### 🤔 What's changed?

The parallel (worker-thread) runtime now settles the "run test cases" phase immediately when there are no test cases to run, instead of waiting forever for a `FINISHED` event that will never arrive.

Concretely, `TestCasesPhase` (`src/runtime/parallel/test_cases_phase.ts`) now resolves its promise as soon as it's constructed with an empty queue:

```ts
this.queue.push(...assembledTestCases)
if (this.queue.length === 0) {
  this.resolve(true)
}
```

I've also added `src/runtime/parallel/test_cases_phase_spec.ts` with unit tests for `TestCasesPhase` in isolation (no worker threads/IPC needed), including a regression test that reproduces the hang by asserting the promise settles instead of timing out.

When running with `parallel > 0` and a `tags`/`paths` selection that matches **zero scenarios**, `cucumber-js` hangs indefinitely instead of exiting - it never prints a summary and never terminates, so any CI job running it just runs until the job-level timeout kills it.

Root cause: `WorkerThreadsAdapter.startPhase()` asks each worker for a command via `phase.fill()`. When the queue is empty, `TestCasesPhase.fill()`/`select()` returns `undefined` for every worker, so **no command is ever issued**. The phase's promise is only ever resolved from `next()`, which is only invoked in response to a `FINISHED` event from a worker (`WorkerThreadsAdapter.handleEventFromWorker`). Since no command was issued, no `FINISHED` event ever arrives, so `resolve`/`reject` are never called and the run hangs forever.

This is a regression introduced by the parallel runtime rewrite in 13.0.0 (#2710, worker-thread based coordinator). I confirmed this does **not** happen on 12.9.0 (pre-rewrite), and does happen on both 13.0.0 and the latest 13.2.0.

Reproduction:
see https://github.com/cucumber/cucumber-js/issues/2907

### 🤖 Disclosure

This change (root-cause analysis, fix, and tests) was investigated and drafted with the help of GitHub Copilot. The hang was reproduced locally, the fix was validated against the reproduction case and the existing/added test suites (all passing), and everything above was reviewed by me before opening this PR.

### 📋 Checklist:

----


src/runtime/parallel/test_cases_phase.ts
```diff
@@ -18,6 +18,11 @@ export class TestCasesPhase implements Phase<RunTestCaseCommand> {
     assembledTestCases: ReadonlyArray<AssembledTestCase>
   ) {
     this.queue.push(...assembledTestCases)
```
davidjgoss: Pushing to the queue is redundant - we can just check the length of `assembledTestCases` and resolve at that point?

davidjgoss (commented):
Nice catch and thanks for raising this.

For testing, we prefer to work at the acceptance level here to keep the implementation flexible. Could you add a scenario to `features/parallel.feature` that covers the zero-scenarios case and ensure it fails without your changes and passes with them? That should be the only test we need.

src/runtime/parallel/test_cases_phase.ts
```diff
@@ -18,6 +18,11 @@ export class TestCasesPhase implements Phase<RunTestCaseCommand> {
...
+    // If there's nothing to run, no worker will ever be issued a command, so no FINISHED
+    // event will ever arrive to trigger settlement via `next()` - settle right away instead.
```
hbl18, line 21: @davidjgoss do you want me to keep this comment in here?

davidjgoss: Yep that's fine.

davidjgoss (approved):

